PR #12, opened 2023-05-11
Add kratos rock release

nsklikas:
Add files for the [Kratos ROCK](github.com/canonical/kratos-rock/)

Didn't add any documentation or contacts, hope that's ok.

### DIFF
--- a/oci/kratos/builds.yaml
+++ b/oci/kratos/builds.yaml
@@ -1,0 +1,9 @@
+version: 1
+
+images:
+  - source: "canonical/kratos-rock"
+    commit: 07ba17388852851a2b731d9b01f55f95271572d6
+    directory: .
+    release-to:
+      risks:
+        - edge
--- a/oci/kratos/builds.yaml
+++ b/oci/kratos/builds.yaml
@@ -4,6 +4,3 @@ images:
   - source: "canonical/kratos-rock"
     commit: 07ba17388852851a2b731d9b01f55f95271572d6
     directory: .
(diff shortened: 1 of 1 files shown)


cjdcordeiro:
Thanks for this contribution. This seems to be a new ROCK and you're asking for it to be released into DH and ECR, so before even starting the review, I suggest you go through the Contributor's Contract (find the link from last week's workshop) and reach out to @valentincanonical about whether this should be under the "ubuntu" namespace or not

oci/kratos/builds.yaml
```diff
@@ -0,0 +1,9 @@
+version: 1
+
+images:
+  - source: "canonical/kratos-rock"
+    commit: 07ba17388852851a2b731d9b01f55f95271572d6
+    directory: .
+    release-to:
+      risks:
+        - edge
```
valentincanonical, line 9:
```suggestion
```

valentincanonical (changes requested):
As discussed, this image might eventually make it to the Ubuntu namespace as part of Canonical's images portfolio but this will happen later when there's better testing, documentation, and the team is ready to commit to supporting these images. 

For now, I suggest removing the release-to / releases options, so that the OCI factory will still build these Rocks but not publish them anywhere beyond the staging GHCR area.

cjdcordeiro:
Hi again! I'm just reviving this thread to know we it should be resumed or closed ;)

(FYI @nsklikas )

PR #34, opened 2023-07-27
Charmed OpenSearch - bump version to 2.9.0

Mehdi-Bendriss:
This PR essentially bumps the OpenSearch version to `2.9.0` 
===> some CVEs are expected to be fixed by this upgrade

Question: How does this play out with versions? I would like the previous version `2.8.0` to also be published


cjdcordeiro (approved):
Ack. This will replace the unpublished 2.8.0 ;)

PR #55, opened 2023-09-06
Run tests on pr

cjdcordeiro:

---

## Description
Every time we build an image, test it too! That includes on pushes and PRs.

### Related issues
Up to now, we have only been testing images after the PRs get merged and before uploading them. This created cases where the image builds well but then fails the tests, causing us to have to redo the PRs.

---


.github/workflows/Image.yaml
```diff
@@ -202,17 +203,83 @@ jobs:
...
+      - name: Enforce test conclusion
+        if: ${{ steps.run-tests.outputs.run-conclusion != 'success' }}
+        # The previous step doesn't always raise an error
+        run: |
+          url='${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ steps.run-tests.outputs.run-id }}'
+          echo "Testing of image '${{ matrix.oci-image }}' failed at [${url}](${url})."
+          exit 1
+
+      # Wait to make sure the test artifacts are uploaded
+      - id: wait-for-artifacts
```
SamirPS, line 248: are you sure sleep 5 is a good pratice for this kind of things ? Can't we get the id of the test action and check the result ?

SamirPS (changes requested):

.github/workflows/Image.yaml
cjdcordeiro: we have the ID of the run, but that's not the problem. In fact, in the previous step, we even wait for the workflow to finish. However, I've noticed in past runs that GitHub might conclude a run before the artefacts are fully available.

And we don't have an artefact ID to wait for, so this is the ugliest but simplest solution

SamirPS (approved):

.github/workflows/Image.yaml
```diff
@@ -27,6 +27,7 @@ on:
 env:
   YQ_VERSION: "v4.32.2"
   YQ_BINARY: yq_linux_amd64
```
jnsgruk, line 29: Side point that's not very important but saves these details:

- Generally I think `snap install yq` is likely to be the right route
- I **think** `yq` is already installed on the Github Actions runners ;-)

SamirPS: For the second point, I think it would be better if we install the software ourselves to ensure we have the correct version of 'yq'.

cjdcordeiro: Oh, indeed it already has `yq` installed (v4.35.1). Thanks for the tip. Being addressed in https://github.com/canonical/oci-factory/pull/56

PR #57, opened 2023-09-08
feat: create chiselled Nginx ROCK

cjdcordeiro:

---

## Description
Add an internal chiselled Nginx ROCK based on the existing ubuntu/nginx. Unlike the existing one, this is a chiselled ROCK and not a Docker image.

It will be kept internal (GHCR only) for now, since we can't yet build the ROCKs for multiple architectures.


oci/nginx/image.yaml
```diff
@@ -0,0 +1,6 @@
+version: 1
+
+upload:  
+  - source: https://git.launchpad.net/~ubuntu-docker-images/ubuntu-docker-images/+git/nginx
```
SamirPS, line 4: can't we use more ssh url?

cjdcordeiro: more? like more than 1? Why?

SamirPS: oh no like instead of https use git+ssh

cjdcordeiro: but then you'd need to add credentials to the OCI factory

SamirPS: yeah true

SamirPS (approved):

PR #119, opened 2024-02-02
Rocks 973/dont update rock if eol is reached

linostar:

---

## Description

The OCI Factory currently auto-updates a ROCK whenever there is a new base ubuntu image upstream.

This task involves using the end-of-life field to infer whether the rock should be updated or not.

## Acceptance Criteria

* PR is rejected if eol is in the past
* Vulnerability scan is stopped if eol is reached
* Auto update is stopped if eol is reached

### DIFF
--- a/oci/mock-rock/_releases.json
+++ b/oci/mock-rock/_releases.json
@@ -1,32 +1,32 @@
 {
     "latest": {
         "candidate": {
-            "target": "151"
+            "target": "1.0-22.04_candidate"
         },
         "beta": {
-            "target": "151"
+            "target": "latest_candidate"
         },
         "edge": {
-            "target": "151"
+            "target": "latest_beta"
         }
     },
     "1.0-22.04": {
         "candidate": {
-            "target": "151"
+            "target": "154"
         },
         "beta": {
-            "target": "151"
+            "target": "154"
         },
         "edge": {
-            "target": "151"
+            "target": "154"
         }
     },
     "test": {
         "beta": {
-            "target": "151"
+            "target": "1.0-22.04_beta"
         },
         "edge": {
-            "target": "151"
+            "target": "test_beta"
         }
     }
 }
--- a/src/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
+++ b/src/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
@@ -15,7 +15,10 @@ import io
 import json
 import logging
 import os
+from datetime import datetime
+
 import requests
+from image.utils.schema.triggers import ImageSchema
 import swiftclient
 import sys
 import tempfile
@@ -65,6 +68,17 @@ def find_released_revisions(releases_json: dict) -> dict:
     return released_revisions
 
 
+def get_image_eol(image_schema: ImageSchema) -> dict:
+    """Given an image schema, returns the eol for each version"""
+    eol_dict = {}
+
+    if image_schema.upload and image_schema.upload[0].release:
+        for version, value in image_schema.upload[0].release.items():
+            eol_dict[version] = value.end_of_life
+
+    return eol_dict
+
+
 # Get the new Ubuntu release from the CLI args
 ubuntu_release = str(sys.argv[1])
 logging.info(f"Ubuntu release: {ubuntu_release}")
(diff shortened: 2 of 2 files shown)


.github/workflows/Vulnerability-Scan.yaml
```diff
@@ -43,20 +44,33 @@ jobs:
...
       - uses: actions/checkout@v3
 
+      - name: Validate end-of-life field values
+        run: |
+          CURRENT_DATE=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
+          EOL_DATES=$(grep -oP 'end-of-life: "\K[0-9-]+T[0-9:]+Z' ${{ inputs.oci-image-path }}/image.yaml)
+
+          while IFS= read -r EOL_DATE; do
+            if [[ "$EOL_DATE" < "$CURRENT_DATE" ]]; then
+              SKIP_VULN_TEST=1
```
cjdcordeiro, line 54: I haven't tested it, but I'm almost sure this won't work. This is a local variable and won't affect the other scopes. You want an output variable instead

cjdcordeiro (changes requested):
I actually think you can solve all of this with a single line change, but tell me if I'm wrong:

why not, in https://github.com/canonical/oci-factory/blob/main/src/image/utils/schema/triggers.py#L18-L20, add:

```python
@validator("end_of_life")
    def ensure_still_supported(cls, v):
        if v < datetime.isoformat(datetime.utcnow()):
            raise ImageReachedEol("....")
        return v
```

Since all the workflows rely on Image.yaml, which relies on this schema, with some exception handling, you can build the logic to know when to trigger something or not

linostar:
> I actually think you can solve all of this with a single line change, but tell me if I'm wrong:
> 
> why not, in https://github.com/canonical/oci-factory/blob/main/src/image/utils/schema/triggers.py#L18-L20, add:
> 
> ```python
> @validator("end_of_life")
>     def ensure_still_supported(cls, v):
>         if v < datetime.isoformat(datetime.utcnow()):
>             raise ImageReachedEol("....")
>         return v
> ```
> 
> Since all the workflows rely on Image.yaml, which relies on this schema, with some exception handling, you can build the logic to know when to trigger something or not

That should work for PR-Validator and _Auto-updates workflows, but from what can I see it won't work for Vulnerability-Scan, as the image.yaml file isn't processed or validated there.

src/image/utils/schema/triggers.py
```diff
@@ -23,6 +27,14 @@ class ImageUploadReleaseSchema(pydantic.BaseModel):
     class Config:
         extra = pydantic.Extra.forbid
 
+    @pydantic.validator("end_of_life")
+    @classmethod
+    def ensure_still_supported(cls, v: datetime) -> datetime:
+        """ensure that the end of life isn't reached."""
+        if v < datetime.now(timezone.utc):
+            raise ImageReachedEol("This image revision has reached its end of life")
```
cjdcordeiro, line 35:
```suggestion
            raise ImageReachedEol("This track has reached its end of life")
```

cjdcordeiro: what about https://github.com/canonical/oci-factory/pull/119/files#diff-120dfed5520cc4aaf924c8389d168eed6e1dc2e75affe683763c3f1de50e6d86R54?

cjdcordeiro (changes requested):
Do you have a test somewhere?

src/image/utils/schema/triggers.py
linostar: sure thing. I'll add it there too.

src/image/utils/schema/triggers.py
```diff
@@ -28,10 +28,9 @@ class Config:
         extra = pydantic.Extra.forbid
 
     @pydantic.validator("end_of_life")
-    @classmethod
-    def ensure_still_supported(cls, v: datetime) -> datetime:
+    def ensure_still_supported(cls, v: Optional[datetime]) -> Optional[datetime]:
         """ensure that the end of life isn't reached."""
-        if v < datetime.now(timezone.utc):
+        if v is None or v < datetime.now(timezone.utc):
```
cjdcordeiro, line 33: this is a good point. since it's simple, can we force this field to always be a date? (so, no default)

now that you are introducing this verification mechanism, there are no excuses to not provide eol

linostar: in that case, you probably want it to be not Optional anymore. Is there a valid use-case where eol is not required?

linostar: @cjdcordeiro we will need to update mock-rock's release/test to include an end-of-life in that case.

cjdcordeiro: sounds good

linostar: @cjdcordeiro All done.

oci/mock-rock/image.yaml
```diff
@@ -5,6 +5,7 @@ release:
     end-of-life: "2024-05-01T00:00:00Z"
     candidate: 1.0-22.04_candidate
   test:
+    end-of-life: "2024-05-01T00:00:00Z"
```
cjdcordeiro, line 8:
```suggestion
    end-of-life: "2026-05-01T00:00:00Z"
```

otherwise we have to change it again in 3 months

cjdcordeiro (changes requested):

cjdcordeiro (approved):
TY

PR #123, opened 2024-02-08
make github workflows callable instead of using workflow-dispatch-action

linostar:

---

## Description

Currently, workflows in `Documentation.yaml`, `Image.yaml` and `Release.yaml` are dispatched via `mathze/workflow-dispatch-action`. We’ll still still need the dispatch on some, at least for those that the Temporal workflows will trigger. But internally (on PRs and commits), those workflows are to be made callable.

### DIFF
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -61,7 +61,7 @@ jobs:
       - name: Infer number of image triggers
         uses: tj-actions/changed-files@v35
         id: changed-files
-        if: github.event_name != 'workflow_dispatch'
+        if: github.event_name != 'workflow_dispatch' && github.event_name != 'workflow_call'
         with:
           separator: ","
           dir_names: "true"
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -73,8 +73,11 @@ jobs:
         run: |
           set -ex
 
-          if [[ "${{ github.event_name }}" != "workflow_dispatch" ]]
-          then
+          case "${{ github.event_name }}" in
+          workflow_dispatch|workflow_call)
+            img_path="oci/${{ inputs.oci-image-name }}"
+            ;;
+          *)
             img_path="${{ steps.changed-files.outputs.all_changed_files }}"
             occurrences="${img_path//[^,]}"
             if [ ${#occurrences} -ne 0 ]
@@ -82,9 +85,8 @@ jobs:
               echo "ERR: can only build 1 image at a time, but trying to trigger ${img_path}"
               exit 1
             fi
-          else
-            img_path="oci/${{ inputs.oci-image-name }}"
-          fi
+            ;;
+          esac
           test -d "${img_path}"
 
           echo "img-name=$(basename ${img_path})" >> "$GITHUB_OUTPUT"
@@ -471,10 +473,10 @@ jobs:
 
   release:
     name: Release
-    needs: [prepare-releases]
+    needs: [prepare-build, prepare-releases]
     uses: ./.github/workflows/Release.yaml
     with:
-      oci-image-name: "${{ inputs.oci-image-name }}"
+      oci-image-name: "${{ needs.prepare-build.outputs.oci-img-name }}"
       image-trigger-cache-key: "${{ github.run_id }}-image-trigger"
     secrets: inherit
 
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -472,7 +472,7 @@ jobs:
   release:
     name: Release
     needs: [prepare-releases]
-    uses: canonical/oci-factory/.github/workflows/Release.yaml@main
+    uses: ./.github/workflows/Release.yaml
     with:
       oci-image-name: "${{ needs.prepare-build.outputs.oci-img-name }}"
       image-trigger-cache-key: "${{ github.run_id }}-image-trigger"
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -188,7 +188,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix: ${{ fromJSON(needs.prepare-build.outputs.build-matrix) }}
-    uses: canonical/oci-factory/.github/workflows/Tests.yaml@main
+    uses: ./.github/workflows/Tests.yaml
     with:
       oci-image-name: "${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.revision }}"
       oci-image-path: "oci/${{ matrix.name }}"
@@ -409,8 +409,8 @@ jobs:
   # We need this job because we want to make the releases all in one go,
   # so that we know which revisions to release, and so that we can update
   # and commit the _releases.json file in a single commit, outside a matrix job
-  dispatch-releases:
-    name: Dispatch releases
+  prepare-releases:
+    name: Prepare releases
     needs: [prepare-build, upload]
     runs-on: ubuntu-22.04
     if: ${{ needs.prepare-build.outputs.release-to != '' }}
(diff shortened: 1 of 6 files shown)


cjdcordeiro (changes requested):

.github/workflows/Image.yaml
```diff
@@ -460,20 +470,11 @@ jobs:
           key: ${{ github.run_id }}-image-trigger
 
       - name: Dispatch Releases workflow
-        # Using this action because others can have this problem:
-        # https://github.com/convictional/trigger-workflow-and-wait/issues/61
-        uses: mathze/workflow-dispatch-action@v1.2.0
+        uses: canonical/oci-factory/.github/workflows/Release.yaml@main
```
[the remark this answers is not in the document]
linostar: @cjdcordeiro  I isolated the workflow calls so they are not steps anymore. Please check and let me know what you think.
I also made some testing on another forked branch. I could trigger running Image.yaml workflow by introducing some changes in mock-rock/image.yaml, but the upload and later jobs are skipped:
https://github.com/canonical/oci-factory/actions/runs/7835687598/job/21381735279

.github/workflows/Image.yaml
```diff
@@ -459,22 +469,26 @@ jobs:
...
+    name: Write dispatch releases summary
+    needs: [dispatch-releases-workflow]
+    runs-on: ubuntu-22.04
+    if: ${{ needs.prepare-build.outputs.release-to != '' }}
+    concurrency:
+      group: ${{ needs.prepare-build.outputs.oci-img-path }}
+      cancel-in-progress: true
+    env:
+      REVISION_DATA_DIR: revision-data
+    steps:
```
cjdcordeiro, line 491: none of this is needed anymore. it was only here because of the previous dispatch

.github/workflows/Image.yaml
```diff
@@ -459,22 +469,26 @@ jobs:
...
-          token: ${{ secrets.GITHUB_TOKEN }}
-          ref: ${{ github.ref_name }}
-          fail-on-error: true
-          workflow-name: Release.yaml
-          payload: '{ "oci-image-name": "${{ needs.prepare-build.outputs.oci-img-name }}", "image-trigger-cache-key": "${{ github.run_id }}-image-trigger"}'
-          run-id: dummy
-          trigger-timeout: "10m"
-          wait-timeout: "20m"
-          use-marker-step: true
+  dispatch-releases-workflow:
```
cjdcordeiro, line 472:
```suggestion
  release:
```

the name of the previous job should also be changed -> "prepare-releases"

.github/workflows/Release.yaml
```diff
@@ -138,25 +148,19 @@ jobs:
...
+    with:
+      oci-image-name: "${{ inputs.oci-image-name }}"
+    secrets: inherit
+
 
+  write-documentation-dispatch-summary:
+    runs-on: ubuntu-22.04
+    name: Write documentation dispatch summary
+    needs: [dispatch-documentation]
+    steps:
```
cjdcordeiro, line 163: same as above. no need for summary or URLs cause the job will run as part of the same run_id - ie. same workflow execution

.github/workflows/_Test-OCI-Factory.yaml
```diff
@@ -11,28 +11,24 @@ on:
...
+    strategy:
+      matrix:
+        oci-image: [mock-rock]
+    uses: canonical/oci-factory/.github/workflows/Image.yaml@main
+    with:
+      oci-image-name: "${{ matrix.oci-image }}"
+      upload: true
+    secrets: inherit
+
+  enforce-conclusion:
```
cjdcordeiro, line 24: same - not needed

.github/workflows/Image.yaml
```diff
@@ -459,22 +469,26 @@ jobs:
...
+    name: Dispatch Releases workflow
+    needs: [dispatch-releases]
+    uses: canonical/oci-factory/.github/workflows/Release.yaml@main
```
cjdcordeiro, line 475:
```suggestion
    uses: ./.github/workflows/Release.yaml
```

otherwise you can't test it outside this repo and the `main` branch.

cjdcordeiro: same for all others calls below

cjdcordeiro: e.g.: https://github.com/canonical/oci-factory/actions/runs/7843366977

cjdcordeiro (changes requested):

.github/workflows/Image.yaml
cjdcordeiro: still using `main`

cjdcordeiro (changes requested):

cjdcordeiro (changes requested):
It's not passing: https://github.com/canonical/oci-factory/actions/runs/7875730443/job/21488453710

linostar:
> It's not passing: https://github.com/canonical/oci-factory/actions/runs/7875730443/job/21488453710

Passing now: https://github.com/canonical/oci-factory/actions/runs/7884195596/job/21512794174

.github/workflows/Image.yaml
```diff
@@ -459,34 +469,14 @@ jobs:
...
-        run: |
-          url='${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ steps.run-releases.outputs.run-id }}'
-          echo "Failed to release '${{ needs.prepare-build.outputs.oci-img-name }}' at [${url}](${url})."
-          exit 1
+  release:
+    name: Release
+    needs: [prepare-releases]
+    uses: ./.github/workflows/Release.yaml
+    with:
+      oci-image-name: "${{ inputs.oci-image-name }}"
```
cjdcordeiro, line 477: This will only work when Image.yaml is called, and not on `push` or other events.

cjdcordeiro (changes requested):

.github/workflows/Image.yaml
linostar: @cjdcordeiro It was `${{ needs.prepare-build.outputs.oci-img-name }}` before, but it was strangely giving an empty value, which is very peculiar since previous jobs are able to read that variable value without issue.

cjdcordeiro: @linostar why do you say it is empty? It is being set [here](https://github.com/canonical/oci-factory/actions/runs/7875730443/job/21488246913#step:5:29) (in the previously failed job)

cjdcordeiro: it seems it was not being passed to the called workflow... investigating

cjdcordeiro: Ok I think I might know what it is.

In https://github.com/canonical/oci-factory/blob/d939c480de447ece234fdd15f9b914df36264ba9/.github/workflows/Image.yaml#L76, the inference of that variable is based on whether the workflow has been dispatched or not.

Well, now that you have introduced `workflow_call`, that `if` will be executed (cause the workflow is being **called** and not **disptached**) and that seems to be resulting in those variables being wrongfully set to nothing. 

You need to expand that `if` condition to also encompass the `workflow_call` event

linostar: Should work now:
https://github.com/canonical/oci-factory/actions/runs/7888207148/job/21525424565

I think it was also missing `prepare-build` in `needs` list before, because otherwise it is not able to access `needs.prepare-build.outputs`.

linostar: And it didn't work. Back to investigating...

.github/workflows/Release.yaml
```diff
@@ -138,37 +148,12 @@ jobs:
 
 
   dispatch-documentation:
-    runs-on: ubuntu-22.04
     name: Dispatch documentation
```
cjdcordeiro, line 151:
```suggestion
    name: Update documentation
```

.github/workflows/Release.yaml
```diff
@@ -138,37 +148,12 @@ jobs:
 
 
   dispatch-documentation:
```
cjdcordeiro, line 150:
```suggestion
  update-documentation:
```

cjdcordeiro (commented):
looking good. just cleaning up a few references to "dispatch". Could you please check is there are others?

.github/workflows/Image.yaml
cjdcordeiro: it was an unrelated LP error

linostar:
> looking good. just cleaning up a few references to "dispatch". Could you please check is there are others?

No others found. 

I also added similar checks for "workflow_call" in Documentation.yaml to avoid the same issues when documentation workflow is triggered.

New test run: https://github.com/canonical/oci-factory/actions/runs/7889980323/job/21531378381

cjdcordeiro (approved):
LGTM thanks

PR #125, opened 2024-02-08
feat: add chiselled Python 3.8 rock

rebornplusplus:

---

## Description
This PR adds the oci-factory configurations for a new chiselled Python3.8 rock.

### DIFF
--- a/oci/python/documentation.yaml
+++ b/oci/python/documentation.yaml
@@ -3,8 +3,12 @@ version: 1
 application: Python
 is_chiselled: true
 description: |
-  Python is an interpreted, interactive, object-oriented, open-source
-  programming language.
+  Python is a high-level, general-purpose programming language. Its
+  design philosophy emphasizes code readability with the use of
+  significant indentation. Python is dynamically typed and
+  garbage-collected. It supports multiple programming paradigms,
+  including structured, object-oriented and functional programming.
+  Read more on [python.org](https://www.python.org/).
 
 debug:
   text: |
--- a/oci/python/documentation.yaml
+++ b/oci/python/documentation.yaml
@@ -1,0 +1,23 @@
+version: 1
+
+application: Python
+is_chiselled: true
+description: |
+  Python is an interpreted, interactive, object-oriented, open-source
+  programming language.
+
+debug:
+  text: |
+    ### Debugging
+
+    To debug the container:
+
+    ```bash
+    docker logs -f python-container
+    ```
+
+    To get an interactive python3 interpreter:
+
+    ```bash
+    docker exec -it python-container python3
+    ```
--- a/oci/python/documentation.yaml
+++ b/oci/python/documentation.yaml
@@ -10,6 +10,11 @@ description: |
   including structured, object-oriented and functional programming.
   Read more on [python.org](https://www.python.org/).
 
+  This image is a chiselled Ubuntu rock that contains only the Python
+  runtime and its standard libraries. It is production-ready and can
+  also be used as a base if additional application dependencies are
+  needed.
+
 debug:
   text: |
     ### Debugging
--- a/oci/python/image.yaml
+++ b/oci/python/image.yaml
@@ -1,0 +1,6 @@
+version: 1
+
+upload:
+  - source: canonical/chiselled-python
+    commit: a9fb59b6066cbe479ee2c395e316094b96af22b3
+    directory: python3.8/
--- a/oci/python/image.yaml
+++ b/oci/python/image.yaml
@@ -9,11 +9,3 @@ upload:
         end-of-life: "2025-01-26T00:00:00Z"
         risks:
           - edge
-      3.8-20.04:
-        end-of-life: "2025-01-26T00:00:00Z"
-        risks:
-          - edge
-      3-20.04:
-        end-of-life: "2025-01-26T00:00:00Z"
-        risks:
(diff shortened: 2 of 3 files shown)


oci/python/image.yaml
```diff
@@ -0,0 +1,6 @@
+version: 1
+
+upload:
+  - source: canonical/chiselled-python
+    commit: a9fb59b6066cbe479ee2c395e316094b96af22b3
+    directory: python3.8/
```
cjdcordeiro, line 6: why isn't it being published?

oci/python/documentation.yaml
```diff
@@ -0,0 +1,23 @@
+version: 1
+
+application: Python
+is_chiselled: true
+description: |
+  Python is an interpreted, interactive, object-oriented, open-source
+  programming language.
```
cjdcordeiro, line 7: This plagiarizes the upstream's description. Let's change it

cjdcordeiro (changes requested):

oci/python/documentation.yaml
rebornplusplus: Updated in https://github.com/canonical/oci-factory/pull/125/commits/41549c18435d3e992b8c779fe9a2071b25c0b415.

oci/python/image.yaml
rebornplusplus: I am sorry, what do you mean by not published?

cjdcordeiro: You're missing the `release` field. Without it, this rock will be built but only kept in GHCR:

https://github.com/canonical/oci-factory/blob/796c8d44e61944de00b08f0de4701192d0ab8e8d/README.md?plain=1#L154-L157

rebornplusplus: Updated in https://github.com/canonical/oci-factory/pull/125/commits/14fcd910957dc1ecf911b8472fea24fb52b8980f. Please have a look.

oci/python/documentation.yaml
```diff
@@ -3,8 +3,12 @@ version: 1
...
 is_chiselled: true
 description: |
-  Python is an interpreted, interactive, object-oriented, open-source
-  programming language.
+  Python is a high-level, general-purpose programming language. Its
+  design philosophy emphasizes code readability with the use of
+  significant indentation. Python is dynamically typed and
+  garbage-collected. It supports multiple programming paradigms,
+  including structured, object-oriented and functional programming.
+  Read more on [python.org](https://www.python.org/).
```
cjdcordeiro, line 11:
```suggestion
  Read more on [python.org](https://www.python.org/).
  
  This image is a chiselled Ubuntu rock that contains only the Python runtime and
  its standard libraries. It is production-ready and can also be used as a base if
  additional application dependencies are needed.
```

oci/python/image.yaml
```diff
@@ -2,5 +2,18 @@ version: 1
...
 upload:
   - source: canonical/chiselled-python
-    commit: a9fb59b6066cbe479ee2c395e316094b96af22b3
+    commit: 3ce55f41e0e33ce1db3552fb18e8179c4a29941a
     directory: python3.8/
+    release:
+      3.8.10-20.04:
+        end-of-life: "2025-01-26T00:00:00Z"
+        risks:
+          - edge
```
cjdcordeiro, line 11:
```suggestion
```

oci/python/image.yaml
```diff
@@ -2,5 +2,18 @@ version: 1
...
+      3.8-20.04:
+        end-of-life: "2025-01-26T00:00:00Z"
+        risks:
+          - edge
+      3-20.04:
+        end-of-life: "2025-01-26T00:00:00Z"
+        risks:
+          - edge
```
cjdcordeiro, line 19:
```suggestion
```

Don't get me wrong, you were doing the right thing :smiley: 

It's just that channels aren't that great to manage in the OCI Factory, since everything is descriptive and going through PRs. In an ideal scenario, your channels were super accurate.

But IMO we should avoid this complexity for now, cause if tomorrow we introduce 3.9, we'll need to remember to manually set the `3-20.04` for that version instead. 

Just playing it safe for now.

cjdcordeiro (changes requested):

oci/python/image.yaml
rebornplusplus: Got it, thanks. Added more commits.

oci/python/image.yaml
```diff
@@ -0,0 +1,11 @@
+version: 1
+
+upload:
+  - source: canonical/chiselled-python
+    commit: 3ce55f41e0e33ce1db3552fb18e8179c4a29941a
+    directory: python3.8/
+    release:
+      3.8.10-20.04:
```
cjdcordeiro, line 8:
```suggestion
      3.8-20.04:
```

I think you've removed the wrong one. It's less error prone to just merge the suggestions

cjdcordeiro (changes requested):

oci/python/image.yaml
rebornplusplus: Indeed, yeah, sorry about that. Updated.

rebornplusplus:
Sorry about the force-push above. The last automatic commit through github suggestion got authored by my personal email which fails CLA check. The diff stays the same. Only the author of the last commit has changed.

PR #128, opened 2024-02-13
fix: replace workflow call/dispatch check with more reliable tests

linostar:

---

## Description
Follow up on merged PR#123, since _Tests workflow didn't seem to pass after the merge.

This changes the check for worflow dispatch/call from relying on `github.event_name` to check whether `inputs.oci-image-name` exists, as the latter is always provided by a workflow dispatch/call.

The issue arised from the fact the the called workflows are inherting the event name from their calling parents, so `github.event_name` isn't a fit anymore for the current purpose.

### DIFF
--- a/.github/workflows/Documentation.yaml
+++ b/.github/workflows/Documentation.yaml
@@ -51,11 +51,12 @@ jobs:
         run: |
           set -ex
 
-          case "${{ github.event_name }}" in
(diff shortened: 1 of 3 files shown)


cjdcordeiro (approved):
great, thanks for the quick fix!

PR #130, opened 2024-02-19
Add end-of-life field to _releases.json

linostar:

---

## Description
Field `end-of-life` is needed in `_releases.json` file so auto generated image trigger objects are complete and don't throw an exception when being generated.

### Related issues
PR https://github.com/canonical/oci-factory/pull/119


linostar:
@cjdcordeiro This should be ready once the tests finish.
You can see the output of the new releases.json in the `Release` step of _Test OCI Factory workflow.

linostar:
New _.releases.json:
https://github.com/canonical/oci-factory/commit/c78aadc2dabfe6347af20c72ecdb703385e45ed3

src/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
```diff
@@ -200,6 +200,11 @@ def find_released_revisions(releases_json: dict) -> dict:
                     release_to[str(to_track)] = {"risks": [to_risk]}
                 else:
                     release_to[to_track]["risks"].append(to_risk)
+                # add end-of-life field to each track
+                if releases[to_track].get("end-of-life"):
```
cjdcordeiro, line 204: but shouldn't we be checking that the image is still supported? if it has reached its EOL, we shouldn't be updating it.

Similarly for the Continuous-Testing, we should be using the new _release.json fields to skip images that no longer should be scanned

cjdcordeiro (changes requested):

src/image/release.py
```diff
@@ -75,7 +76,7 @@
         all_releases[track] = {}
 
     for risk, value in risks.dict(exclude_none=True).items():
-        if risk == "end-of-life":
+        if risk in ["end-of-life", "end_of_life"]:
```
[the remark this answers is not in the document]
linostar: Always at the moment. As a result of that, "end_of_life" is considered as an "unknown risk" and skipped altogether. The mechanism of adding eol values to the releases dict is already there before this PR, but this value has been always skipped because the field comes as "end_of_life".

src/tests/get_released_revisions.py
```diff
@@ -84,6 +85,11 @@ def get_image_name_in_registry(img_name: str, revision: str) -> str:
 
         released_revisions[img] = []
         for risks in releases.values():
+            if risks.get("end-of-life") and risks["end-of-life"] < datetime.now(
+                timezone.utc
+            ).strftime("%Y-%m-%dT%H:%M:%SZ"):
```
cjdcordeiro, line 90: please log something, so that we're not kept in the dark when this occurs

src/tests/get_released_revisions.py
```diff
@@ -84,6 +85,11 @@ def get_image_name_in_registry(img_name: str, revision: str) -> str:
 
         released_revisions[img] = []
         for risks in releases.values():
+            if risks.get("end-of-life") and risks["end-of-life"] < datetime.now(
```
cjdcordeiro, line 88: well...this is a tricky one. if there's no EOL, we could skip indeed. But I think...we shouldn't. I.e. if we don't know the EOL, the safest is to simply keep supporting it. There should be a warning for it, ofc, but it would be dismissive and possibly dangerous to skip it

src/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
```diff
@@ -205,6 +206,10 @@ def find_released_revisions(releases_json: dict) -> list:
                     release_to[to_track]["end-of-life"] = releases[to_track][
                         "end-of-life"
                     ]
+                    if release_to[to_track]["end-of-life"] < datetime.now(
+                        timezone.utc
+                    ).strftime("%Y-%m-%dT%H:%M:%SZ"):
+                        del release_to[to_track]
```
cjdcordeiro, line 212: this is over complicating it. Just

```
if releases[to_track]["end-of-life"] < datetime.now...:
    # skip it, with logs

release_to[to_track]["end-of-life"] = releases[to_track]["end-of-life"] 
```

so, no need to `del`, and as above, you shouldn't update it EOL has passed. The logic should be exactly the same as for other jobs

cjdcordeiro (changes requested):

src/tests/get_released_revisions.py
linostar: @cjdcordeiro  This will skip only if there is an EOL value and its older than current datetime.

src/tests/get_released_revisions.py
```diff
@@ -84,10 +84,13 @@ def get_image_name_in_registry(img_name: str, revision: str) -> str:
             releases = json.load(rf)
 
         released_revisions[img] = []
-        for risks in releases.values():
+        for revision, risks in releases.items():
```
cjdcordeiro, line 87:
```suggestion
        for track, risks in releases.items():
```

it's the track, not the revision (same below in the logs)

src/tests/get_released_revisions.py
cjdcordeiro: yes you're right. bad reading on my side.

can you pls include a warning log for cases where the eol is not set?

src/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
```diff
@@ -201,15 +201,20 @@ def find_released_revisions(releases_json: dict) -> list:
                     release_to[str(to_track)] = {"risks": [to_risk]}
                 else:
                     release_to[to_track]["risks"].append(to_risk)
+
                 # add end-of-life field to each track
                 if releases[to_track].get("end-of-life"):
-                    release_to[to_track]["end-of-life"] = releases[to_track][
-                        "end-of-life"
-                    ]
                     if release_to[to_track]["end-of-life"] < datetime.now(
```
cjdcordeiro, line 207: variable unassigned

src/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
```diff
@@ -201,15 +201,20 @@ def find_released_revisions(releases_json: dict) -> list:
...
                         timezone.utc
                     ).strftime("%Y-%m-%dT%H:%M:%SZ"):
-                        del release_to[to_track]
+                        logging.info(
+                            f"Skipping revision {to_track} because it reached "
```
cjdcordeiro, line 211:
```suggestion
                            f"Skipping track {to_track} because it reached "
```

src/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
```diff
@@ -201,15 +201,20 @@ def find_released_revisions(releases_json: dict) -> list:
...
+                            "its end of life"
```
cjdcordeiro, line 212: pls add the EOL date while logging

cjdcordeiro (approved):
ty

PR #132, opened 2024-02-28
fix: extract revision numbers only from the revision dicts

linostar:

---

## Description
Fix issue in https://github.com/canonical/oci-factory/actions/runs/8073081089/job/22056099446 caused by eol value being misused as a revisions dict.

### DIFF
--- a/oci/mock-rock/_releases.json
+++ b/oci/mock-rock/_releases.json
@@ -13,13 +13,13 @@
     },
     "1.0-22.04": {
         "candidate": {
-            "target": "185"
+            "target": "186"
         },
         "beta": {
-            "target": "185"
+            "target": "186"
         },
         "edge": {
-            "target": "185"
+            "target": "186"
         },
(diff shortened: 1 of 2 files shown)


cjdcordeiro (changes requested):
this is not gonna fix it. the proposed change is redundant and is actually introducing an error, that is `revision` may not be defined

linostar:
> this is not gonna fix it. the proposed change is redundant and is actually introducing an error, that is `revision` may not be defined

@cjdcordeiro I handled the case where `revision` is not defined. But why do you say the proposed change is redundant? The issue originates from the fact that `targets` can be either a revision dict or a eol datetime string, so this makes it extracts the revision number only in the former case.

linostar:
@cjdcordeiro done.

cjdcordeiro (approved):

cjdcordeiro:
Thanks

PR #133, opened 2024-03-01
fix: add end-of-life fields to old releases in their respective _releases.json

linostar:

---

## Description

This is to allow old oci images without end-of-life fields to build properly until they get a new revision.


cjdcordeiro (changes requested):
Unfortunately it needs to be a bit more complex than this.

Requirements:
  1. we want all new PRs to have an EOL field in the trigger file
  2. we will allow existing rocks to have their trigger/_release.json files without an EOL

because of 1., the schema fields must not be optional.
But then, because of 2. we must allow them to be optional. 

Clunky, I know.

I think the real problem with https://github.com/canonical/oci-factory/actions/runs/8099754626/job/22136244098 is that it was **disptached** even though its EOL were not defined.

Here's the problem: https://github.com/canonical/oci-factory/actions/runs/8099737147/job/22136206712#step:6:872

See how it is dispatching rebuilds, without including an EOL for each track

linostar:
@cjdcordeiro 

The implications currently go beyond _releases.json, as the _Test-OCI-Factory workflow call Image workflow, which turns load the oci image file into an image trigger. So if we don't want to make the eol field optional in the ImageSchema model, I guess we will have to have another model that can used in the except block?  Note that the ImageSchema model is used in 3 different files called by different workflows.

If I am not missing anything, the ways I can think of to make current oci images with missing eol fields still build:
- add a valid eol value to them (like 1 year in the future) in the code after loading the yaml but before passing the yaml data to the ImageSchema model constructor
- run eol checks outside Pydantic, i.e. eol fields are back to being optional, and extra checks are added after the model is validated, but only for new PRs
- ? 

Let me know what you think.

cjdcordeiro:
> @cjdcordeiro
> 
> The implications currently go beyond _releases.json, as the _Test-OCI-Factory workflow call Image workflow, which turns load the oci image file into an image trigger. So if we don't want to make the eol field optional in the ImageSchema model, I guess we will have to have another model that can used in the except block? Note that the ImageSchema model is used in 3 different files called by different workflows.
> 

I don't understand this. Why is `_Test-OCI-Factory` a problem? This only runs for internal purposes and whenever the mock-rock or our CI scripts change. The called Image.yaml will read the image trigger from the `mock-rock` which **must contain** the EOL fields -> it's our test rock after all

> If I am not missing anything, the ways I can think of to make current oci images with missing eol fields still build:
> 
> * add a valid eol value to them (like 1 year in the future) in the code after loading the yaml but before passing the yaml data to the ImageSchema model constructor
> * run eol checks outside Pydantic, i.e. eol fields are back to being optional, and extra checks are added after the model is validated, but only for new PRs
> * ?
> 
> Let me know what you think.

I might be missing something, but IMO the issue is in https://github.com/canonical/oci-factory/actions/runs/8099737147/job/22136206712#step:6:872 which should be including the EOL fields in the payload, and isn't

cjdcordeiro (changes requested):
I see some channels still without an end-of-life. is there a way we can get those from history (git tags)?

linostar:
> I see some channels still without an end-of-life. is there a way we can get those from history (git tags)?

@cjdcordeiro I wanted to ask you about those. Given that the remaining tracks all have old eol values, do you need to be accurate here? I mean if the purpose of _releases.json is just to provide boolean decision on whether a particular track is going to be built/tested, then we can put an old eol value for all those tracks, no?

cjdcordeiro:
better safe than sorry. if it is too convoluted to get those dates, then at least double check with the images' maintainers (they are pretty much all owned by the same team anyway :p)

cjdcordeiro (approved):
thanks @linostar !

@lucabello we're now starting to enforce EOL for all rocks, could you please double check if this backport of the EOLs into each `_releases.json` file is accurate?

lucabello:
We don't really care about the precise EOL date anyway, we just set it about a year in the future, so I'd say the changes look good to me!

PR #134, opened 2024-03-06
feat: add support for arm64 runners and builds

cjdcordeiro:

---

## Description
There are many commits in this PR, but they all contribute to the same:
 - finer control over when our internal tests run
 - get rid of nodejs warnings by bumping many 3rd party actions
 - make the build of rocks a reusable workflow
 - add support for our self-hosted ARM64 runners
 - use arm64 runners whenever the rock recipe asks for it

### DIFF
--- a/.github/workflows/_Test-OCI-Factory.yaml
+++ b/.github/workflows/_Test-OCI-Factory.yaml
@@ -1,10 +1,15 @@
-name: _Test OCI Factory
+name: _Test OCI Factory | mock-rock
 on:
-  push:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - master
     paths:
       - ".github/workflows/*"
       - "!.github/workflows/CLA-Check.yaml"
       - "!.github/workflows/PR-Validator.yaml"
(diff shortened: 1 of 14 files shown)


.github/workflows/_Test-OCI-Factory.yaml
```diff
@@ -1,22 +1,24 @@
-name: _Test OCI Factory
+name: _Test OCI Factory | mock-rock
 on:
-  push:
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - master
```
linostar, line 6:
```suggestion
      - main
```

cjdcordeiro: :facepalm: thanks

linostar (changes requested):
It seems that after the changes in Release workflow, the `git pull --quiet` action is failing as the tracked branch is missing (see https://github.com/canonical/oci-factory/actions/runs/8185830180/job/22385117117?pr=134)

linostar (approved):
LGTM, thanks!

PR #135, opened 2024-03-07
fix: race condition between Image and _Test

cjdcordeiro:

---

## Description
 - bumped the download-artifact action in the provenance job (which was failing to fetch the workflow's artifacts)
 - fixed a race condition: in the previous PR, the Image.yaml and _Test workflows would be triggered at the same time, which would cause a race condition to the `_release.json` file. This PR makes sure the Image.yaml doesn't run for the mock-rock, cause it should be called by the _Test workflow anyway

### Related issues
https://github.com/canonical/oci-factory/actions/runs/8187580397/job/22390383561

### DIFF
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -450,7 +450,7 @@ jobs:
     needs: [upload]
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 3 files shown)


linostar (approved):
nice and good.

PR #139, opened 2024-03-13
fix(ci): matrix-generated outputs are useless

cjdcordeiro:

---

## Description
The code is currently using the output from a reusable workflow to know the cache key where to fetch the OCI archive from.
Unfortunately, this only works for single builds - i.e., if multiple rocks are being built in parallel, in a matrix, this output get overwritten by the last cell of the matrix, and thus the subsequent jobs relying on  that output variable will be using the wrong value.

### Related issues
Fixes this error: https://github.com/canonical/oci-factory/actions/runs/8263457709/job/22605630636?pr=138

---


cjdcordeiro:
Need to merge this cause there are many builds failing. pls shout if u see something wrong

linostar (commented):
Looks fine to me. If anything is still going wrong we will make a follow-up. Thanks!

PR #140, opened 2024-03-13
[DPE-3726] Fix ZooKeeper version

deusebio:

---

## Description
We updated the version of the ZooKeeper rock to fix the CVEs, but we did not update the track. Sorry for this oversight!


oci/zookeeper/image.yaml
```diff
@@ -5,7 +5,7 @@ upload:
     commit: 1781da1dacbacf647319731ff25ba4ad00082140
     directory: .
     release:
-      3.8.4-22.04:
+      3.8-22.04:
```
deusebio, line 8: It is probably better to skip the patch version

cjdcordeiro: btw, now it is already published. How do you want to handle it?

deusebio: can we remove it? I suppose we should since it is also wrong. The branch has a name that does not correspond to the actual workload version

cjdcordeiro: this can be tricky. since it's a recently pushed tag, we can try and hope there are no complaints on the user side.

For that, please update https://github.com/canonical/oci-factory/blob/main/oci/zookeeper/_releases.json#L2-L7. I'll take care of removing the tags on the registry side

deusebio: Yes, I **really** don't want this to be an habit. I don't think there should be any problem to do it in a specific case, given it is a super super fresh version and track. I agree with you if the image had already multiple pulls. BTW, can we check the pulls of a specific hash in dockerhub as much as we can do in GHCR?

cjdcordeiro: through parsing of CSV files, yes

cjdcordeiro (approved):
let's give it a try. thanks

PR #148, opened 2024-03-14
fix(ci): uniquely identify rocks from parallel builds

cjdcordeiro:

---

## Description

With the recent introduction of arm64 support, the rocks are being built in a matrix and then assembled into a multi-arch rock. The assembling relies on the existing artifacts uploaded during the matrix build, but it is currently downloading all of them and not filtering them correctly - the result is a multi-arch roch with multiple manifests for the same arch.

This PR fixes that by pretending a unique identifier to each single-arch rock artifact.


linostar (approved):
LGTM, thanks!

PR #153, opened 2024-04-02
feat: add Node.js 20 rock

linostar:

---

## Description

This PR adds the oci-factory configurations for Node.js runtime rock.


cjdcordeiro:
We are not yet publishing Node rocks, and we should especially be careful about such a Rock because we wouldn't be able to offer proper support (since Node20 isn't in the archives). For 24.04, Node18 is our target, but we'll keep that on hold for now, at least until we have a more tangible opportunity to pursue.

PR #160, opened 2024-04-10
feat: add lpci support for multi arch rocks

cjdcordeiro:

---

## Description

Solely for rocks that target architectures other than amd64 and arm64, this PR introduces new CI steps to add support for lpci-based remote builds, via Launchpad.

NOTE: the lp-build is still somewhat unstable and thus this is a sort of try and error build, with many retries. It should only be used for rocks that really need the multi arch. 

Example of a successful run: https://github.com/canonical/oci-factory/actions/runs/8618748045
Image in DH: https://hub.docker.com/r/rocksdev/mock-rock/tags?page=&page_size=&ordering=&name=1.2

### Related issues

---

*Picture of a cool rock:*

image

### DIFF
--- a/.github/workflows/Build-Rock.yaml
+++ b/.github/workflows/Build-Rock.yaml
@@ -151,10 +151,10 @@ jobs:
             cd ${{ inputs.rockfile-directory }}
             rocks_toolbox="$(mktemp -d)"
             git clone --depth 1 --branch v1.1.2 https://github.com/canonical/rocks-toolbox $rocks_toolbox
-            ${rocks_toolbox}/rocks-toolbox/rockcraft_lpci_build/requirements.sh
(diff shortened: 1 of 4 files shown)


linostar (commented):
LGTM. Thanks.
I would suggest to build with lpci only the architectures that are not supported in github runners. So for the mock-rock example, build amd64 with github and ppc64el with lpci. This at least should decrease the errors we get from lpci builds, while we find a better alternative!

cjdcordeiro:
Thanks. That could indeed be a future improvement. It wouldn't work now because the `lpci` script is designed to work like `rockcraft remote-build`, which means it will solely rely on the `platforms` field inside the `rockcraft.yaml` file. So it would always end up building both `amd64` and `ppc64el`.

linostar (approved):

PR #167, opened 2024-04-22
Parse numerical strings in yaml files as strings

linostar:

---

## Description
The default python YAML parser that we uses to parse various yaml files consider all numeric values in the yaml values as integers or float as long as they are not enclosed by quotes.
That means that a field like `version: 2.3` would get the value of `2.3` as a float and not as string.
However, our python classes are expecting all values to be strings, because those values are not guaranteed to be numerical in all cases (i.e. `version` value can contain letters too).

## Action taken
The implemented solution in this PR was:
- add a common module and import it in all other modules that uses the yaml parser
- in the common module, make the default parsing behavior of numerical values construct them as strings

### DIFF
--- a/oci/mock-rock/_releases.json
+++ b/oci/mock-rock/_releases.json
@@ -1,68 +1,68 @@
 {
     "latest": {
         "candidate": {
-            "target": "226"
+            "target": "1.0-22.04_candidate"
         },
         "beta": {
-            "target": "226"
+            "target": "latest_candidate"
         },
         "edge": {
-            "target": "226"
+            "target": "latest_beta"
         },
         "end-of-life": "2025-05-01T00:00:00Z"
     },
     "1.0-22.04": {
         "candidate": {
-            "target": "226"
+            "target": "247"
         },
         "beta": {
-            "target": "226"
+            "target": "247"
         },
         "edge": {
-            "target": "226"
+            "target": "247"
         },
         "end-of-life": "2025-05-01T00:00:00Z"
     },
     "test": {
         "beta": {
-            "target": "226"
+            "target": "1.0-22.04_beta"
         },
         "edge": {
-            "target": "226"
+            "target": "test_beta"
         },
         "end-of-life": "2026-05-01T00:00:00Z"
     },
     "1.1-22.04": {
         "end-of-life": "2025-05-01T00:00:00Z",
         "candidate": {
-            "target": "227"
+            "target": "248"
         },
         "beta": {
-            "target": "227"
+            "target": "248"
         },
         "edge": {
-            "target": "227"
+            "target": "248"
         }
     },
     "1-22.04": {
         "end-of-life": "2025-05-01T00:00:00Z",
         "candidate": {
-            "target": "227"
+            "target": "248"
         },
         "beta": {
-            "target": "227"
+            "target": "248"
         },
         "edge": {
-            "target": "227"
+            "target": "248"
         }
     },
     "1.2-22.04": {
         "end-of-life": "2025-05-01T00:00:00Z",
         "beta": {
-            "target": "228"
+            "target": "249"
         },
         "edge": {
-            "target": "228"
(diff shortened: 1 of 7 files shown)


SamirPS (commented):
lgtm

.github/workflows/Build-Rock.yaml
```diff
@@ -72,6 +72,20 @@ jobs:
...
+                "tag:yaml.org,2002:int",
+                string_constructor,
+                Loader=yaml.SafeLoader
+            )
+            yaml.add_constructor(
+                "tag:yaml.org,2002:float",
+                string_constructor,
+                Loader=yaml.SafeLoader
+            )
+
```
cjdcordeiro, line 88: why do we need this here? afaiu this code is only using the `platforms` field, which can never be a number.

linostar: Safe precaution for the future. If you deem it unncessary I'll remove it.

cjdcordeiro: let's remove it cause it's not doing anything but bloating an already fragile piece of code

cjdcordeiro (changes requested):

src/image/utils/custom_yaml.py
```diff
@@ -0,0 +1,14 @@
+import yaml
+
+
+def string_constructor(loader, node):
+    """change node type to string"""
+    return loader.construct_scalar(node)
+
+
+yaml.add_constructor(
+    "tag:yaml.org,2002:int", string_constructor, Loader=yaml.SafeLoader
```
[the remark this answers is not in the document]
linostar: Because next time someone writes a new yaml.load in the code, they may not be aware (or they may forget) that they should use different loader. Hence I opted for the approach, so that all current (and futuristic) yaml loads will use the custom string constrcutor without any further modification on yaml method calls (aka yaml.safe_load).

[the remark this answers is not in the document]
cjdcordeiro: that is something a unit test should solve (which is yet something we're missing).

I'd rather keep it simple. This approach is introducing a lot of relative imports and unused imports.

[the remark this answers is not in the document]
linostar: done

src/uploads/infer_image_track.py
```diff
@@ -4,8 +4,10 @@
 import os
 import logging
 import subprocess
+import sys
```
cjdcordeiro, line 7: unused

src/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
```diff
@@ -25,6 +25,18 @@
...
+    return loader.construct_scalar(node)
+
+
+yaml.add_constructor(
+    "tag:yaml.org,2002:int", string_constructor, Loader=yaml.SafeLoader
+)
+yaml.add_constructor(
+    "tag:yaml.org,2002:float", string_constructor, Loader=yaml.SafeLoader
+)
+
```
cjdcordeiro, line 39: why do we need to keep this one?

src/image/utils/custom_yaml.py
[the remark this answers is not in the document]
cjdcordeiro: this can be deleted now

oci/mock-rock/_releases.json
cjdcordeiro, line 1: can u pls resolve this conflict

cjdcordeiro (changes requested):

oci/mock-rock/_releases.json
linostar: Each time I try to resolve conflicts using github web UI, checks start to fail because it merge other commits from main as well. I'll have to do it manually.

cjdcordeiro (approved):
All good now. thanks for the changes

PR #194, opened 2024-06-10
fix: remove trigger for img w/o release within eol

zhijie-yang:

---

## Description
Fix the problem of an image with all releases EOL having a trigger with no `end-of-life` field. 

E.g., https://github.com/canonical/oci-factory/actions/runs/9415158556 and https://github.com/canonical/oci-factory/actions/runs/9415158867.

Such images will not be triggered to build since this commit.

### DIFF
--- a/src/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
+++ b/src/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
@@ -221,10 +221,12 @@ if __name__ == "__main__":
                             f"Track {to_track} is missing its end-of-" "life field"
                         )
 
+                # filter out the tracks with releases out of EOL
+                release_to = {k : v for k, v in release_to.items() if "end-of-life" in v}
+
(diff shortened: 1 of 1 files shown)


src/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
```diff
@@ -221,10 +221,12 @@ def find_released_revisions(releases_json: dict) -> list:
                             f"Track {to_track} is missing its end-of-" "life field"
                         )
 
+                # filter out the tracks with releases out of EOL
+                release_to = {k : v for k, v in release_to.items() if "end-of-life" in v}
```
linostar, line 225: The code is correct, but I'd rather not loop over the release_to dict again unnecessarily. Instead, you can add an `else` block to the inner `if` block above that check for eol against current date, and delete `to_track` key from `release_to` dict there.

linostar (changes requested):

PR #198, opened 2024-06-11
The CLI Client

zhijie-yang:

---

## Description
This pull request implements a cli client for OCI Factory, such that it is no longer required to create pull requests to trigger image builds and releases for the images.

For details see specification RK040.


src/cli-client/README.md
```diff
@@ -0,0 +1,40 @@
+
+# The OCI Factory CLI Client
+
+A CLI client that triggers GitHub workflows for building, uploading and releasing images in [OCI Factory](https://github.com/canonical/oci-factory).
+
+## How to use
+
+### New User Onboarding
```
SamirPS, line 8: I think you can remove the onboarding word

src/cli-client/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+See ["How to Contribute as a Maintainer"](https://github.com/canonical/oci-factory?tab=readme-ov-file#as-a-maintainer--).
+
+Upon finishing the onboarding, the user should receive a GitHub Personal Access Token. This token grants the user proper
```
SamirPS, line 12: "grants you" would be better no?

src/cli-client/README.md
```diff
@@ -0,0 +1,40 @@
...
+permissions to trigger the build, upload and release the Rocks. This token should never be shared with third-parties, nor
```
SamirPS, line 13: of your Rock ?

SamirPS (commented):
I will let someone else review the go code

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,57 @@
...
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/canonical/go-flags"
+	"github.com/canonical/oci-factory/cli-client/internals/logger"
+)
+
+// Example: oci-factory upload -y --release tracks=1.0-22.04,risks=edge,eol=2025-06-01
```
letFunny, line 13: I think comments like this in the code not attached to anything can easily get outdated. If you want to include examples it may be better to do it in the CLI help which will also be more useful for the end user.

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,57 @@
...
+
+type CmdRelease struct {
+	Revision int `long:"revision" description:"The integer-typed revision number of the image release"`
```
letFunny, line 16: [Nit]: I dont have enough context about the flag but I think `integer-typed` is not necessary here, probably `revision number of the image release` is enough together with the long help for the command if there needs to be any additional consideration.

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+var opts struct {
+	// Leave `release` here for now. Should be moved into a `cmd_release.go` in phase 2.
```
letFunny, line 20: I would add a `TODO` here so that you can easily grep for it afterwards.

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,57 @@
...
+		fmt.Printf("%s [Y/n]: ", s)
+
+		res, err := r.ReadString('\n')
+		if err != nil {
+			logger.Panicf("%v", err)
+		}
+
+		res = strings.TrimSpace(res)
+
+		if res == "y" || res == "yes" || res == "" {
```
letFunny, line 48: UX-wise I think `res == ""` should mean no, at least I am used to that when running command like `rm`.

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,57 @@
...
+			return
+		} else if res == "n" || res == "no" {
+			fmt.Println("Cancelled")
+			os.Exit(0)
```
letFunny, line 52: Should this be a return code != 0 to indicate that we didn't perform the operation, again looking at commands like `rm` it seems to do that.

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,57 @@
...
+
+func blockForConfirm(s string, tries int) {
+	r := bufio.NewReader(os.Stdin)
+
+	for ; tries > 0; tries-- {
+		fmt.Printf("%s [Y/n]: ", s)
+
+		res, err := r.ReadString('\n')
+		if err != nil {
+			logger.Panicf("%v", err)
```
letFunny, line 43: Is this going to produce an stack trace if the user does something like CTRL-C? Because if that is the case I think it would be better to just return error code != 0 and maybe log the error.

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,108 @@
...
+
+type CmdUpload struct {
+	UploadRelease []string `long:"release" description:"Release images to container registries.\nSyntax: --release tracks=<release track>,risks=<risk1>[,<risk2>...],eol=yyyy-mm-dd"`
+}
+
+var cmdUpload CmdUpload
+var releases = make([]UploadRelease, 0)
+
+func init() {
+	// fmt.Println("DEBUG Adding upload to parser commands")
```
letFunny, line 26: These statements should not be part of the PR, either remove them or use `debugf` or something similar if you want to log something for debugging purposes.

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,108 @@
...
+	Track     string
+	Risks     []string
+	EndOfLife string
+}
+
+type CmdUpload struct {
+	UploadRelease []string `long:"release" description:"Release images to container registries.\nSyntax: --release tracks=<release track>,risks=<risk1>[,<risk2>...],eol=yyyy-mm-dd"`
+}
+
+var cmdUpload CmdUpload
```
letFunny, line 22: Why is this a global variable? I think you can achieve the same by doing:
```go
parser.Command.AddCommand("upload", "Trigger the build and release for the image in the current working directory", "long description", &CmdUpload{})
```

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,108 @@
...
+var releases = make([]UploadRelease, 0)
```
letFunny, line 23: Why is this a global variable? You can probably accomplish the same by returning the list in `parseUploadReleases` and then using the argument in trigger.

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,108 @@
...
+		parts := strings.Split(argStr, ",")
+
+		var risks []string
+		for _, part := range parts {
+			keyValue := strings.SplitN(part, "=", 2)
+			if len(keyValue) == 1 && len(risks) > 0 {
+				// This is part of the risks list (e.g., "stable" in "risks=beta,stable")
+				risks = append(risks, keyValue[0])
+				continue
+			}
```
letFunny, line 51: I dont have enough context about the parser but this looks like it is trying to catch the case where `risks=foo,bar` is at the end of the string. I have two suggestions here but I am not sure about them because I don't know the full details. I would recommend:
1. Having a comment describing the format.
2. Thinking of a way of making the parser a little bit more uniform. It does not have to be more complex necessarily but the format is a little bit ambiguous (commas and equals) so probably you need something a little bit more sophisticated. That can be regular expression or even a different way of splitting the string ( cannot think of good way now without regular expression though).

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,108 @@
...
+				fmt.Errorf("unknown key: %s", key)
+			}
+		}
+		release.Risks = risks
+		releases = append(releases, release)
+	}
+	return nil
+}
+
+// use the global variable `releases` directly
```
letFunny, line 75: Comments in go are prefixed with the function name and are written in a different way normally. See https://go.dev/doc/comment#func for examples and for more information.

src/cli-client/internals/cli/validator.go
```diff
@@ -0,0 +1,18 @@
+package cli
+
+import (
+	"fmt"
+	"time"
+)
+
+func ValidateAndFormatDate(dateStr string) (string, error) {
```
letFunny, line 8: I may lack some context here but if this function doesn't need to be public and it is only public to test it I think it is better to use the `export_test.go` trick or to think of way of testing this through the more general function.

src/cli-client/internals/cli/validator_test.go
```diff
@@ -0,0 +1,29 @@
...
+func TestValidateAndFormatDateBadInput(t *testing.T) {
+	inputs := []string{"2001-02-29", "2024-05-32", "01-01-2023"}
+
+	for _, datetime := range inputs {
+		_, err := cli.ValidateAndFormatDate(datetime)
+		if err == nil {
+			t.Fatalf("Parsing illegal date input [%s] didn't raise error", datetime)
+		}
+	}
+}
```
letFunny, line 29: This depends on the style but normally these two tests are combined into the same table that has an input and an error field. Then the testing code can do:
```go
for _, test := range testCases {
	_, err := cli.ValidateAndFormatDate(test.input)
	if test.err != nil {
		c.Assert(err, Equals, test.err)
		break
	}
	c.Assert(err, IsNil)
}
```

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,108 @@
...
+func triggerUploadReleases() {
+	// really builds
+	if len(releases) > 0 {
```
letFunny, line 78: In Go we normally prefer early returns instead of conditionals, so this function could do:
```go
if len(releases) == 0 {
    fmt.Println("Nothing passed in --release, upload will be skipped.")
    return
}

buildMetadata := trigger.InferBuildMetadata()
...
```

letFunny (commented):
Skimmed through the first files of the PR, it is looking good but I think it would benefit for following Go's style a little bit more. Left some comments about it.

Also, I think this PR is too big to review (1400 lines) and will also be hard to do rounds if we need to address a lot of things at the same time. This probably needs to be split into several PRs (if possible).

src/cli-client/internals/cli/cli_main.go
zhijie-yang: There won't be a stack trace if the user interrupts the process (per my tests). The process will just exit with code 130.

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: Makes sense.

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: Nice suggestion. Fixed.

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: Good suggestion. Fixed.

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: Good catch. Fixed.

src/cli-client/internals/cli/cli_main.go
zhijie-yang: Good catch. Changed to exit 1.

src/cli-client/internals/cli/cli_main.go
zhijie-yang: Fixed.

src/cli-client/internals/cli/cli_main.go
zhijie-yang: Makes sense!

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: Good suggestion. Modified and will be used in the future comments.

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: Good suggestion with the format comment. 

For point 2, since we don't force the user input to be of a fixed order of `tracks`, `risks` and `eol`, a regular expression might not be a good idea. Therefore I think using an FSM-styled parser would better suit the need here.

src/cli-client/internals/cli/validator.go
zhijie-yang: Good suggestion. Added `export_test.go` to cli package.

src/cli-client/README.md
zhijie-yang: All nice suggestions. Fixed.

src/cli-client/internals/cli/cli_upload.go
letFunny: I don't have a strong opinion about regular expressions, if there is something simpler then that is indeed much better. However, just for the sake of completeness, you could do it with a regular expression with something like:
```
(((tracks=[^,]+)|(risks=([,^,]+,)*[^,]+)|(eol=[^,]+)),){0,2}(((tracks=[^,]+)|(risks=([,^,]+,)*[^,]+)|(eol=[^,]+)))
```
and then do some validation to see that there is only one `risks`, one `tracks`, etc.

Note: you can simplify the regex a lot by referring to previous capture group using `(?1)` and things like that.

zhijie-yang: Tried out a bit. Would you think 
```
tracks=([^,]+),risks=([^,]+(?:,[^,]+)*),eol=(\d{4}-\d{2}-\d{2})
```
is a better idea?

zhijie-yang: Would you think the following patter better?
```
(((tracks=[^,]+)|(risks=[^,]+,*[^,]+)|(eol=[^,]+)),?){3}
```

letFunny: The problem with that one is that it matches an input with a trailing comma at the end, right?

src/cli-client/internals/cli/validator_test.go
letFunny: I am sorry I wrote the code above with a `break` instead of a `continue`, that would be ideal (even though assert fails the test, if you ever change it to `c.Check` instead).

src/cli-client/internals/cli/cli_upload.go
letFunny: But let's not use a regular expression if there is something simpler and the parsing code is easier to follow.

thp-canonical (commented):
Added some comments, hope it's useful.

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+type UploadRelease struct {
+	Track     string
+	Risks     []string
+	EndOfLife string
+}
+
+type CmdUpload struct {
+	UploadRelease []string `long:"release" description:"Release images to container registries.\nSyntax: --release tracks=<release track>,risks=<risk1>[,<risk2>...],eol=yyyy-mm-dd"`
```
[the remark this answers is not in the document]
zhijie-yang: @cjdcordeiro Since we are only releasing to one track with one call of `--release`, should we change it to "track"? (Currently in our spec it's still "tracks".)

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,121 @@
...
+	return nil
+}
+
+// parseUploadReleases parses the release arguments and returns a list of UploadRelease structs.
+// The release arguments are expected to be in the format "tracks=<track>,risks=<risk1>[,<risk2>...],eol=yyyy-mm-dd".
+// Multiple release arguments can be passed, separated by spaces.
+func parseUploadReleases(args []string) ([]UploadRelease, error) {
+	var release UploadRelease
+	releases := make([]UploadRelease, 0)
+	regex := regexp.MustCompile("(((tracks=[^,]+)|(risks=[^,]+,*[^,]+)|(eol=[^,]+)),?){3}")
```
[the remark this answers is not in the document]
zhijie-yang: We are not currently enforcing the ordering of the keys from the user input. @cjdcordeiro Shall we enforce that? If this is enforced, then using regex to extract the inputs would become much easier.

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,55 @@
...
+
+	for ; tries > 0; tries-- {
+		fmt.Printf("%s [y/N]: ", s)
+
+		res, err := r.ReadString('\n')
+		if err != nil {
+			logger.Panicf("%v", err)
+		}
+
+		res = strings.TrimSpace(res)
```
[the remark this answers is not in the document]
zhijie-yang: Good one, fixed!

src/cli-client/internals/cli/cli_upload.go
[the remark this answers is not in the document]
cjdcordeiro: It should be `track`. Which means we might end up some repetition like `--release track=foo,risks=edge,beta,eol=01-2025 --release track=foo-alias,risks=edge,beta,eol=01-2025`. But that's ok for now. We can later introduce a new option `track-aliases` such that one could have `--release track=foo,risks=edge,beta,eol=01-2025,track-aliases=foo-alias,foo2...`

src/cli-client/internals/cli/cli_upload.go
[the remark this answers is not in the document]
zhijie-yang: Or maybe, I can match only a key-value pair of the string each time, using `tracks=([^,]+),?` to match tracks, `risks=(((beta|candidate|edge|stable),?)+),?` to match risks and `eol=(\d{4}-\d{2}-\d{2}),?` to match eol separately?

This could mitigate the complexity of implementing such a regex, and also provide finer-grained error messages for the user.

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return releases, nil
+}
+
+func triggerUploadReleases(releases []UploadRelease) {
+	// really builds
+	if len(releases) == 0 {
+		fmt.Println("No release track specified, no build will be triggered.")
+		return
+	}
```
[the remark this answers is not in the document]
zhijie-yang: Good catch. Fixed

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,55 @@
...
+		parser.WriteHelp(os.Stdout)
+		os.Exit(0)
+	}
+	parser.Parse()
+}
+
+func blockForConfirm(s string, tries int) {
+	r := bufio.NewReader(os.Stdin)
+
+	for ; tries > 0; tries-- {
```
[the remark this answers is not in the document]
zhijie-yang: Nice suggestion. I eliminated the whole loop instead. This loop can be just unnecessary.

src/cli-client/internals/client/wf_poller.go
```diff
@@ -0,0 +1,189 @@
...
+	return GetWorkflowJobsProgressFromResp(responseBody)
+}
+
+// TODO how to implement test for calls into GetWorkflowRunID?
+func WorkflowPolling(workflowExtRefId string) {
+	runId, _ := GetWorkflowRunID(workflowExtRefId)
+	logger.Debugf("%d\n", runId)
+
+	fmt.Printf("Task %s started. Details available at %s%d.\n", workflowExtRefId,
+		"https://github.com/canonical/oci-factory/actions/runs/", runId)
```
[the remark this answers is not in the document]
zhijie-yang: Good catch. Fixed.

src/cli-client/internals/cli/cli_upload_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+func TestParseUploadReleases(t *testing.T) {
+	for _, test := range []struct {
+		in     string
+		out    []cli.UploadRelease
+		errMsg string
+		err    bool
+	}{
+		{
+			"tracks=1.0.0,risks=high,low,eol=2023-01-01",
```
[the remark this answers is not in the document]
zhijie-yang: This is discussed and negated in the spec.

src/cli-client/internals/cli/cli_upload.go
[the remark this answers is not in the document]
zhijie-yang: Refactored and noted in ROCKS-1233.

src/cli-client/internals/client/export_test.go
```diff
@@ -0,0 +1,13 @@
...
+	GetWorkflowRunID                = getWorkflowRunID
+	GetWorkflowRunStatusFromResp    = getWorkflowRunStatusFromResp
+	GetWorkflowJobsProgressFromResp = getWorkflowJobsProgressFromResp
+)
+
+func SetWorkflowDispatchURL(url string) string {
+	ret := workflowDispatchURL
+	workflowDispatchURL = url
+	return ret
+}
```
thp-canonical, line 13: One nice pattern from Pebble (not sure if it's a general Go pattern) is that the returned value will be a callable, so you don't need to know how to restore it, [example from Pebble](https://github.com/canonical/pebble/blob/d822e9fd9c484a9c35e1b8f08b8d438ca2115225/internals/overlord/state/task.go#L413-L416) and [its usage](https://github.com/canonical/pebble/blob/d822e9fd9c484a9c35e1b8f08b8d438ca2115225/internals/overlord/state/task_test.go#L357-L358):

```suggestion
func SetWorkflowDispatchURL(url string) (restore func()) {
	orig := workflowDispatchURL
	workflowDispatchURL = url
	return func() { workflowDispatchURL = orig }
}
```

src/cli-client/internals/client/wf_dispatcher_test.go
```diff
@@ -0,0 +1,71 @@
...
+	defer mockServer.Close()
+
+	request, _ := http.NewRequest(http.MethodPost, mockServer.URL, bytes.NewBuffer(mockJson))
+	client.SetHeaderWithMap(request, header)
+
+	// Call the DispatchWorkflow function
+	originalURL := client.SetWorkflowDispatchURL(mockServer.URL)
+	client.DispatchWorkflow(expectedPayload)
+	token.RestoreTokenEnv(saveToken)
+	client.SetWorkflowDispatchURL(originalURL)
```
thp-canonical, line 70: With the [restore function pattern](https://github.com/canonical/pebble/blob/d822e9fd9c484a9c35e1b8f08b8d438ca2115225/internals/overlord/standby/standby_test.go#L128) this could become:

```suggestion
	// Call the DispatchWorkflow function
	defer client.SetWorkflowDispatchURL(mockServer.URL)()
	client.DispatchWorkflow(expectedPayload)
	token.RestoreTokenEnv(saveToken)
```

Note that `client.SetWorkflowDispatchURL(mockServer.URL)` is evaluated immediately (and it returns a `func()`), but the call to the returned func (that's the `()` at the end) is deferred, meaning it's called at the end of the scope, which is especially nice if there are multiple exit points).

Possibly clearer / more obvious, but not as compact (should be equivalent to the above):

```suggestion
	// Call the DispatchWorkflow function
	restore := client.SetWorkflowDispatchURL(mockServer.URL)
	defer restore()
	client.DispatchWorkflow(expectedPayload)
	token.RestoreTokenEnv(saveToken)
```

src/cli-client/internals/client/export_test.go
zhijie-yang: Nice one. Fixed.

src/cli-client/internals/cli/cli_upload.go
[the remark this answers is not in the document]
zhijie-yang: Updated in [0c81222](https://github.com/canonical/oci-factory/pull/198/commits/0c81222bcd16b62d8d1ba393a0651fd3e5e584a9).

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: Updated in [0c81222](https://github.com/canonical/oci-factory/pull/198/commits/0c81222bcd16b62d8d1ba393a0651fd3e5e584a9).

src/cli-client/internals/token/token.go
```diff
@@ -0,0 +1,49 @@
...
+
+var _accessToken = ""
+
+func readAccessToken() string {
+	envAccessToken := os.Getenv(TokenVarName)
+	if len(envAccessToken) == 0 {
+		fmt.Print("GitHub personal access token: ")
+		accessTokenBytes, err := term.ReadPassword(0)
+		if err != nil {
+			logger.Panicf("Error handling token: %v", err)
```
linostar, line 21: Just to make sure the error message doesn't contain the token, which will be then displayed in the logs, replace any possible instances of the token inside the error message.

src/cli-client/internals/trigger/export_test.go
```diff
@@ -0,0 +1 @@
+package trigger
```
linostar, line 1: to be removed

linostar (changes requested):

.github/workflows/_Test-OCI-Factory.yaml
```diff
@@ -12,6 +12,7 @@ on:
       - "oci/mock*"
       - "src/**"
       - "!src/workflow-engine/**"
+      - "!src/cli-client/**"
```
cjdcordeiro, line 15: we need a new workflow just to run the unit tests though

src/cli-client/README.md
cjdcordeiro, line 1: this doc can live here, but there must also be at least a reference in the root README.md file, such that people become aware of this. Otherwise, it is buried.

There is already an [whole section about Maintainer contributions](https://github.com/canonical/oci-factory?tab=readme-ov-file#as-a-maintainer--), which must now be updated to describe this 2nd method.

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/canonical/go-flags"
+	"github.com/canonical/oci-factory/cli-client/internals/logger"
+)
+
+type CmdRelease struct {
+	Revision int `long:"revision" description:"The revision number of the image release"`
+}
+
+var opts struct {
+	// TODO: Leave `release` here for now. Should be moved into a `cmd_release.go` in phase 2.
```
cjdcordeiro, line 19: Then I'd remove/comment it. Otherwise people will try to use it (because it's described in the usage info) and fail.

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,59 @@
...
+	CmdRelease       CmdRelease `command:"release" description:"Release (re-tag) the image into the container registries"`
+	SkipConfirmation bool       `short:"y" description:"Skip the confirmation to upload/release an image"`
+}
+
+var parser = flags.NewParser(&opts, flags.Default)
+
+func CliMain() {
+	if len(os.Args) == 1 || os.Args[1] == "help" {
+		parser.WriteHelp(os.Stdout)
```
cjdcordeiro, line 28: We're missing a general description of the `oci-factory` client in the help text.

cjdcordeiro (changes requested):

.github/workflows/_Test-OCI-Factory.yaml
zhijie-yang: Added as ROCKS-1272 in JIRA.

src/cli-client/README.md
zhijie-yang: Added as ROCKS-1273 in JIRA.

src/cli-client/internals/cli/cli_main.go
zhijie-yang: Good catch, fixed.

src/cli-client/internals/cli/cli_main.go
zhijie-yang: Fixed.

linostar (approved):

cjdcordeiro (changes requested):

README.md
```diff
@@ -84,6 +88,23 @@ multiple images;
    - that means one PR can only propose changes to
 [Maintainer files](#maintainer-files) within a single `oci` folder.
 
+#### 2. With the CLI Client
+
+Maintainers can use the OCI Factory's CLI Client to interact with the OCI
+Factory. The CLI Client is a Go module at [src/cli-client](src/cli-client).
+
+Here's what you should do and know before you use the CLI Client to trigger
+the first build of your image:
```
cjdcordeiro, line 97:
```suggestion
Here's what you should do and know before you use the CLI Client:
```

README.md
```diff
@@ -84,6 +88,23 @@ multiple images;
...
+
+1. you are onboard as a Maintainer, and have received your token to the CLI
```
cjdcordeiro, line 99:
```suggestion
1. you **must** already be an onboarded Maintainer and have had significant interactions with the OCI Factory (via PRs). Why is that? To use this client you'll be granted write permissions to the repository, and thus you'll have escalated rights when compared to a regular Maintainer - with great power comes great maintainability,
```

README.md
```diff
@@ -70,7 +70,11 @@ Pull Requests 🚀.
 
 ### As a **Maintainer** ⛏ 🪨
 
-Maintainers shall use Pull Requests (PRs) as an interface for asking the OCI
+There are two ways to use the OCI Factory to build the images.
```
cjdcordeiro, line 73:
```suggestion
Maintainers can request new image builds and releases by either creating Pull Requests (PRs) or via the OCI Factory CLI Client.
```

src/cli-client/snap/snapcraft.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+parts:
+  oci-factory:
+    plugin: go
+    build-snaps: [go/latest/stable]
+    source: https://github.com/canonical/oci-factory
+    source-type: git
+    source-subdir: src/cli-client/cmd/oci-factory
+    stage-packages:
+      - git
```
cjdcordeiro, line 20: why is this needed?

src/cli-client/snap/snapcraft.yaml
```diff
@@ -0,0 +1,31 @@
...
+    # TODO remove this override-pull when the branch is merged
+    override-pull: |
+      snapcraftctl pull
+      git checkout ROCKS-1216-cli-client
```
cjdcordeiro, line 24: I guess for this :p which I also guess is a leftover?

cjdcordeiro (changes requested):
ty for the changes. just a bunch of nits, and over @rebornplusplus approves it we should be good to merge it

src/cli-client/snap/snapcraft.yaml
zhijie-yang: We need to call the `git rev-parse --show-prefix ` command as a subprocess in the CLI client as the git library does not support `git rev-parse` yet.

src/cli-client/snap/snapcraft.yaml
zhijie-yang: I don't think we already have the CLI client in the default branch of `oci-factory` repo. So I have to add this `override-pull` to test the snapcraft packaging before this branch is merged into main.

src/cli-client/snap/snapcraft.yaml
cjdcordeiro: got it. That means `git` should also be a dependency for those who are installing from source (thus part of the inner README)

src/cli-client/snap/snapcraft.yaml
cjdcordeiro: sure but that's for testing. it shouldn't go to `main` with this branch name

src/cli-client/snap/snapcraft.yaml
zhijie-yang: Yes. I will add it.

src/cli-client/snap/snapcraft.yaml
cjdcordeiro: well, you should actually be using `.` as your source. that way you don't need this.

zhijie-yang: Yeah, good suggestion.

src/cli-client/snap/snapcraft.yaml
zhijie-yang: Other suggested changes on documentation and this is fixed in [49ac137](https://github.com/canonical/oci-factory/pull/198/commits/49ac137eb8fcb59588bf62394ed677cc9f43f7a4).

src/cli-client/README.md
```diff
@@ -18,10 +18,35 @@ The user will be asked to input the GitHub Personal Access Token upon triggering
...
-### Installation
+### Install using Snap
+
+```bash
+sudo snap install oci-factory
+```
+
+### Install using Go
+
+#### Dependencies
```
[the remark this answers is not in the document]
zhijie-yang: I put another header 4 to make it outside of "Dependencies". Fixed in [b7d61f5](https://github.com/canonical/oci-factory/pull/198/commits/b7d61f5b825f754e633319dd01360b2bf81b48da).

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type CmdRelease struct {
+	Revision int `long:"revision" description:"The revision number of the image release"`
+}
+
+const longDescription = `The OCI Factory CLI client is a tool that builds, tests, and releases
+the OCI images owned by Canonical using the Github workflow in 
+the OCI Factory repository.
+`
```
rebornplusplus, line 21: nit: let's format this a bit
```suggestion
const longDescription = `
The OCI Factory CLI client is a tool that builds, tests, and releases the OCI
images owned by Canonical using the Github workflow in the OCI Factory repository.
`
```

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,66 @@
...
+
+var opts struct {
+	// TODO: Leave `release` here for now. Should be moved into a `cmd_release.go` in phase 2.
+	// CmdRelease       CmdRelease `command:"release" description:"Release (re-tag) the image into the container registries"`
+	SkipConfirmation bool `short:"y" description:"Skip the confirmation to upload/release an image"`
+}
+
+var parser = flags.NewParser(&opts, flags.Default)
+
+func CliMain() {
```
rebornplusplus, line 31: We should really do some error checking here. Let's return an error from this func if it does not work.

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,66 @@
...
+	if len(os.Args) == 1 || os.Args[1] == "help" {
+		// TODO: Add subcommand `help` if more complex help messages are needed
+		fmt.Print(longDescription + "\n")
+		parser.WriteHelp(os.Stdout)
+		os.Exit(0)
+	}
+	parser.Parse()
```
rebornplusplus, line 38: We need to check errors. We can use the default example from the author's repo:
```go
	if _, err := parser.Parse(); err != nil {
		switch flagsErr := err.(type) {
		case flags.ErrorType:
			if flagsErr == flags.ErrHelp {
				os.Exit(0)
			}
			os.Exit(1)
		default:
			os.Exit(1)
		}
	}
	return nil
```

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,66 @@
...
+
+var parser = flags.NewParser(&opts, flags.Default)
+
+func CliMain() {
+	if len(os.Args) == 1 || os.Args[1] == "help" {
+		// TODO: Add subcommand `help` if more complex help messages are needed
+		fmt.Print(longDescription + "\n")
+		parser.WriteHelp(os.Stdout)
+		os.Exit(0)
+	}
```
rebornplusplus, line 37: How about the following:
```go
	addHelp(parser)
	if len(os.Args) == 1 {
		parser.WriteHelp(os.Stdout)
		os.Exit(0)
	}
```

And in this file we can add the `addHelp` method:
```go
func addHelp(parser *flags.Parser) {
	parser.ShortDescription = "CLI client to build, test and release OCI images"
	parser.LongDescription = strings.TrimSpace(longDescription)
}
```
We can rename `addHelp` to something more meaningful, like `configParser`.

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,66 @@
...
+	parser.Parse()
+}
+
+func blockForConfirm(s string) {
```
rebornplusplus, line 41: Again, let's return errors from here. We can handle the errors (print them, ..) from the call-site.

src/cli-client/internals/cli/cli_main.go
```diff
@@ -0,0 +1,66 @@
...
+	// check if is a tty
+	fi, err := os.Stdin.Stat()
+	if err != nil || fi.Mode()&os.ModeNamedPipe == 0 {
+		fmt.Fprintln(os.Stderr, "Non-interactive terminal detected. Must run with -y option.")
```
rebornplusplus, line 45: As I was saying, instead of writing the errors here, let's just return those errors and handle them in call-site.

Note that, in this particular case, the program would still continue running since we are not doing anything other than printing.

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+type UploadRelease struct {
+	Track     string
+	Risks     []string
+	EndOfLife string
+}
+
+type CmdUpload struct {
+	UploadRelease []string `long:"release" description:"Release images to container registries.\nSyntax: --release track=<release track>,risks=<risk1>[,<risk2>...],eol=yyyy-mm-dd"`
```
rebornplusplus, line 20: nit: How about `Release` as the struct field?

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+var riskOptions = []string{"stable", "candidate", "beta", "edge"}
+
+func init() {
+	parser.Command.AddCommand("upload", "Trigger the build and release for the image in the current working directory",
```
rebornplusplus, line 26: Why not use `parser.Command` directly?

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,150 @@
...
+		"long description", &CmdUpload{})
```
rebornplusplus, line 27: Need to change the `long description` text!

src/cli-client/internals/cli/cli_upload.go
```diff
@@ -0,0 +1,150 @@
...
+
+		releases = append(releases, release)
+	}
+	if len(releases) == 0 {
+		return nil, fmt.Errorf("no release track specified, no build will be triggered")
+	}
+	return releases, nil
+}
+
+func triggerUploadReleases(releases []UploadRelease) {
```
rebornplusplus, line 126: Need some error returns here. Lots of things within this function can go wrong it seems.

rebornplusplus (changes requested):
An overall go-specific review. Please check if the functionality works or not.

Error handling needs to be a bit better in my opinion. In lots of places, we are trying to handle the errors from within the function by printing or panicking, which although would most likely work, is not a nice way to do it and on the rarest of occasions, might just break. Let's change those up to return said errors from the function.
```go
func foo(args Type) (RetType, error) {
	...
	// return errors whenever something does not work
	...
}

...

val, err := foo(bar)
if err != nil {
	// Handle the error here.
	logger.Panicf("foo does not work: %v", err)
}
```
If, indeed, error-handling within the function is a must, we can do something like this too:
```go
func foo(args Type) (ret RetType, err error) {
	defer func() {
		if err != nil {
			// Handle the error here. Example:
			logger.Panicf(err)
		}
	}()
	...
}
```

src/cli-client/internals/cli/cli_main.go
zhijie-yang: Agree! I think we can keep it as `addHelp` for now as it's not dealing with something more than adding the help messages.

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: What do you mean by using `parser.Command` directly?

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: Fixed in [db12a9a](https://github.com/canonical/oci-factory/pull/198/commits/db12a9a004a5879e6bf79da7e3f0a5d73025c759). All errors are handled internally as exiting or panicing.

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: Do you mean the `Release` from `package trigger`?

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: Fixed in [b819f16](https://github.com/canonical/oci-factory/pull/198/commits/b819f16bca4dbc3ddc1cb659c2d2c255c016c1fe).

src/cli-client/internals/cli/cli_upload.go
rebornplusplus: Sorry `parser.AddCommand`. I must have been dozy!

src/cli-client/internals/cli/cli_upload.go
rebornplusplus: No no, I meant renaming `UploadRelease` to `Release`.

rebornplusplus (approved):
Just checked the go stuffs -- looks OK to me for a first start. Nice work! 

PS. please check whether it works properly or not :) I don't have much context about oci-factory.

src/cli-client/internals/cli/cli_upload.go
zhijie-yang: Nice one! Fixed in [c541687](https://github.com/canonical/oci-factory/pull/198/commits/c5416879a41a4c8447573cb9b9ef207bc73bffc9).

cjdcordeiro (commented):
can u pls ensure all comments are resolved?

cjdcordeiro (approved):
Great! Thanks

PR #201, opened 2024-06-20
Preempt Swift slots to enforce mutual exclusion updating Swift container

zhijie-yang:

---

## Description

This PR adds a `Prepare upload` job to the `Image.yaml` workflow, which preempts the Swift slot right after retrieving the revision number from the Swift container to prevent a race condition. For the workflow before this PR, such race conditions can happen when a workflow is triggered, while another workflow with the same image name runs for a job between `Build` and `Upload` (both included). Since the job `Build` can take up 20 minutes, it's easy to trigger such a workflow that races with other workflows. The eventual consequence of such a race condition can result in duplicated revisions of images of different tracks, thus letting the workflow release wrong artefacts to some tracks (tags).

## Changelog

 * Using `{name}_{commit}_{directory}` as identifiers for OCI image artefacts;
 * Validation of the uniqueness of such identifiers;
 * Forming a critical section in the job `Prepare upload` that makes a write-after-read operation to the Swift container;
 * Preempting slots in Swift container for concurrent uploads in the job `Upload`;
 * Protecting such critical sections by wrapping them with "lock" and "unlock" operations using a lock file (on a per-image basis);
 * Busy waiting when the lock file exists;
 * Unlocking when the workflow is cancelled or the preemption operation fails;
   * Won't unlock if the lock operation fails (usually means waiting for other `Prepare upload` jobs for the same image to finish)
 * Adding a timestamp to the revision data cache key to enable the re-run of failed `Prepare upload` jobs that can give different revision numbers.

### Related issues

 * https://bugs.launchpad.net/ubuntu-docker-images/+bug/2069652
 * https://github.com/canonical/chiselled-python/issues/21

# Before merge:
- [ ] Revert the changes [92e27a6](https://github.com/canonical/oci-factory/pull/201/commits/92e27a65e9c43bb7e9ffe0fe84f3536551a9b701) and [e364915](https://github.com/canonical/oci-factory/pull/201/commits/e3649157196189905809ef9614d1569d1f474e38).

---


.github/workflows/Image.yaml
```diff
@@ -159,22 +144,146 @@ jobs:
...
+      oci-image-name: "${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.dir_identifier }}"
       oci-image-path: "oci/${{ matrix.name }}"
       test-from: "cache"
-      cache-key: ${{ github.run_id }}-${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.revision }}        
+      cache-key: ${{ github.run_id }}-${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.dir_identifier }}        
     secrets: inherit
 
-  upload:
+  
+  prepare-upload:
```
zhijie-yang, line 154: Can move env for the whole job. The risk should be acceptable as this is a simple job.

.github/workflows/Image.yaml
```diff
@@ -159,22 +144,146 @@ jobs:
...
+          DATA_DIR: "revision-data"
+          OS_USERNAME: ${{ secrets.SWIFT_OS_USERNAME }}
+          OS_TENANT_NAME: ${{ secrets.SWIFT_OS_TENANT_NAME }}
+          OS_PASSWORD: ${{ secrets.SWIFT_OS_PASSWORD }}
+          OS_REGION_NAME: ${{ secrets.SWIFT_OS_REGION_NAME }}
+          OS_STORAGE_URL: ${{ secrets.SWIFT_OS_STORAGE_URL }}
+          IMAGE_NAME: ${{ needs.prepare-build.outputs.oci-img-name }}
+          SWIFT_CONTAINER_NAME: ${{ vars.SWIFT_CONTAINER_NAME }}
+        run: |
+          for revision_file in `ls ${{ env.DATA_DIR }}`; do
```
zhijie-yang, line 231: Reduce the times of updating the Swift container. Ideally only once after everything here is ready.

.github/workflows/Image.yaml
```diff
@@ -159,22 +144,146 @@ jobs:
...
+        run: |
+          set -ex
+
+          mkdir ${{ env.DATA_DIR }}
+
+          ./src/image/prepare_single_image_build_matrix.py \
+            --oci-path ${{ needs.prepare-build.outputs.oci-img-path }} \
+            --revision-data-dir ${{ env.DATA_DIR }} \
+            --next-revision ${{ steps.get-next-revision.outputs.revision }} \
+            --infer-image-track
```
zhijie-yang, line 216: Cache the source repos being cloned by passing --infer-image-track for use in the job upload. Detele this cache at the end of the workflow.

zhijie-yang: No need. Inference of track is removed from job upload.

examples/mock-rock/1.2/rockcraft.yaml
```diff
@@ -6,11 +6,11 @@ base: bare
 build-base: ubuntu@22.04
 license: Apache-2.0
 platforms:
-  ppc64el:
+  # ppc64el:
```
linostar, line 9: why is this removed?

src/image/prepare_single_image_build_matrix.py
```diff
@@ -57,16 +63,37 @@ def validate_image_trigger(data: dict) -> None:
...
 
+        if args.infer_image_track:
+            import sys
+            sys.path.append("src/")
+            from git import Repo
+            from tempfile import TemporaryDirectory as tempdir
+            from uploads.infer_image_track import get_base_and_track
+            with tempdir() as d:
+                url = f"https://github.com/{builds[img_number]['source']}.git"
+                repo = Repo.clone_from(url, d)
```
linostar, line 76: add arg `depth=1` to clone_from function to ensure it's a shallow clone.

src/image/requirements.txt
```diff
@@ -1,4 +1,7 @@
+GitPython
 PyYAML
 pydantic==1.9.0
 python-swiftclient
-python-keystoneclient
\ No newline at end of file
+python-keystoneclient
+
```
linostar, line 6: nitpick: no need for an in-between empty line here

src/uploads/infer_image_track.py
```diff
@@ -21,40 +21,47 @@ def get_release_from_codename(codename: str) -> str:
...
-)
-
-try:
-    base_release = float(rock_base.replace(":", "@").split("@")[-1])
-except ValueError:
-    logging.warning(
-        f"Could not infer ROCK's base release from {rock_base}. Trying with codename."
+def get_base_and_track(rockcraft_yaml) -> tuple[str, str]:
+    rock_base = (
+        rockcraft_yaml["base"]
```
linostar, line 26: you should check for build-base first and use it if it's available, otherwise use base.
the reason is the project can have something like build-base: 22.04 and base: 24.04, but the build-base (22.04) is what you want to get.

src/uploads/swift_lockfile_lock.sh
```diff
@@ -0,0 +1,35 @@
...
+pushd "${staging_area}"
+
+# check if the ${IMAGE_NAME}/lockfile.lock exists in the swift container
+# if it does, wait until the timeout is reached and emit an error
+# if it does not, upload the lockfile.lock to the swift container
+# and exit
+while [ $TIMEOUT -gt 0 ]; do
+    swift list $SWIFT_CONTAINER_NAME -p $IMAGE_NAME | grep "lockfile.lock" && sleep $SLEEP_TIME || break
+    TIMEOUT=$(( $TIMEOUT - $SLEEP_TIME ))
+    if [ $TIMEOUT -eq 0 ]; then
```
linostar, line 28: use `$TIMEOUT -lt 1` in case a value that is not multiple of the SLEEP_TIME was used for TIMEOUT.

linostar (changes requested):

examples/mock-rock/1.2/rockcraft.yaml
zhijie-yang: This is to temporarily avoid lpci builds, since they are always failing.

src/uploads/infer_image_track.py
zhijie-yang: This change is just a refactor of the code. This logic is not touched.

src/image/prepare_single_image_build_matrix.py
zhijie-yang: This shouldn't be a shallow clone, at least not recognised by the original implementation.
https://github.com/canonical/oci-factory/blob/8b5294d8bf3e22c1998928df5f8140a0f132d072/.github/workflows/Image.yaml#L200

linostar (approved):

.github/workflows/Image.yaml
```diff
@@ -159,22 +143,118 @@ jobs:
...
+
+          echo "revision-data-cache-key=${{ github.run_id }}-${{ env.DATA_DIR }}-$(date +%s)" >> "$GITHUB_OUTPUT"
```
cjdcordeiro, line 209: this feels off. Does it work? You're creating an unpredictable key. How will the following steps/jobs restore the cache if they don't know the value of `date +%s`?

.github/workflows/Image.yaml
```diff
@@ -159,22 +143,118 @@ jobs:
...
+          ./src/uploads/requirements.sh
+          pip install -r src/image/requirements.txt -r src/uploads/requirements.txt
+
+      - name: Upload the lockfile for the image
+        id: swift-lock
+        run: |
+          ./src/uploads/swift_lockfile_lock.sh \
+            ${{ needs.prepare-build.outputs.oci-img-name }}
+
+      # Here starts the critical section, have to be executed in sequence outside of matrix.
```
cjdcordeiro, line 189: I don't get the comment...what do you mean in "sequence outside of matrix"? this is not a matriced job, and the script has no sequences...it's basically a one-off job

.github/workflows/Image.yaml
```diff
@@ -159,22 +143,118 @@ jobs:
...
+            --oci-path ${{ needs.prepare-build.outputs.oci-img-path }} \
+            --revision-data-dir ${{ env.DATA_DIR }} \
+            --next-revision ${{ steps.get-next-revision.outputs.revision }} \
+            --infer-image-track
+
+          echo "revision-data-cache-key=${{ github.run_id }}-${{ env.DATA_DIR }}-$(date +%s)" >> "$GITHUB_OUTPUT"
+
+      - name: Preempt Swift slot
+        env:
+          DATA_DIR: "revision-data"
```
cjdcordeiro, line 213: why not make this a job env?

.github/workflows/Image.yaml
```diff
@@ -159,22 +143,118 @@ jobs:
...
+        run: |
+          ./src/uploads/preempt_swift_slots.sh ${{ env.DATA_DIR }}
+
+      # Here leaves the critical section.
+      # The lock will be removed even the steps above failed,
```
cjdcordeiro, line 218:
```suggestion
      # The lock will be removed even if the steps above fail,
```

.github/workflows/Image.yaml
```diff
@@ -159,22 +143,118 @@ jobs:
...
+        # However, failing to lock the swift container can mean there are
+        # multiple workflows trying to upload the same image at the same time.
+        # Therefore we should not remove the lockfile if the swift lock failed.
+        if: ${{ always() && steps.swift-lock.outcome != 'failure' }}
+        run: |
+          ./src/uploads/swift_lockfile_unlock.sh \
+            ${{ needs.prepare-build.outputs.oci-img-name }}
+
+      # The revision files have to be sanitised before merging,
+      # since the `track` field should not be present.
```
cjdcordeiro, line 231: why?

src/uploads/swift_lockfile_lock.sh
```diff
@@ -0,0 +1,35 @@
...
+touch "${staging_area}/${IMAGE_NAME}/lockfile.lock"
+
+pushd "${staging_area}"
+
+# check if the ${IMAGE_NAME}/lockfile.lock exists in the swift container
+# if it does, wait until the timeout is reached and emit an error
+# if it does not, upload the lockfile.lock to the swift container
+# and exit
+while [ $TIMEOUT -gt 0 ]; do
+    swift list $SWIFT_CONTAINER_NAME -p $IMAGE_NAME | grep "lockfile.lock" && sleep $SLEEP_TIME || break
```
cjdcordeiro, line 26: can u please add a comment here, to say that

`# There's still the unlikely corner case where 2 concurrent jobs are waiting for the lockfile to get removed, and they may exit the while loop at the same time, getting into a race condition`

cjdcordeiro (changes requested):

.github/workflows/Image.yaml
zhijie-yang: This key is listed as an output of the job `prepare-upload`, so I believe all the following steps/jobs can find it. The motivation to make it unique is to ensure the following runs of a failed workflow run can still put revision data into the cache. Otherwise, the workflow will raise a warning that the cache with the same key exists and is created by a previous workflow run. In this case, the new cache won't be written, and thus cause other errors for the new run.

.github/workflows/Image.yaml
```diff
@@ -159,22 +143,118 @@ jobs:
...
+      OS_STORAGE_URL: ${{ secrets.SWIFT_OS_STORAGE_URL }}
+      IMAGE_NAME: ${{ needs.prepare-build.outputs.oci-img-name }}
+      SWIFT_CONTAINER_NAME: ${{ vars.SWIFT_CONTAINER_NAME }}
+    outputs:
+      build-matrix: ${{ steps.prepare-matrix.outputs.build-matrix }}
+      revision-data-cache-key: ${{ steps.prepare-matrix.outputs.revision-data-cache-key }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Use custom image trigger
```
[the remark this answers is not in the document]
zhijie-yang: These environmental vars are used by all the other steps in this job, except: `checkout`, `setup-python`, installing requirements, `prepare-matrix`, sanitising and caching.

.github/workflows/Image.yaml
zhijie-yang: Fixed.

.github/workflows/Image.yaml
zhijie-yang: Fixed.

.github/workflows/Image.yaml
```diff
@@ -197,38 +277,17 @@ jobs:
...
-      - run: cd source && git checkout ${{ matrix.commit }}
-
       - uses: actions/cache/restore@v4
         with:
           path: ${{ env.OCI_ARCHIVE_NAME }}
-          key: ${{ github.run_id }}-${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.revision }}
+          key: ${{ github.run_id }}-${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.dir_identifier }}
           fail-on-cache-miss: true
 
-      - name: Infer track name
```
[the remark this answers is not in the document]
zhijie-yang: I think this has to be removed from the `upload` job, as the revision number and image tracks should already be inferred in the `prepare-upload` job. Since the only component that needs to clone the image repo is `inter_image_track.py`, and the `infer_image_track.py` is refactored and used as a module in `prepare_` python scripts, just letting the scripts handling their required data would be a better approach iMO.

.github/workflows/Image.yaml
zhijie-yang: The lockfile should be removed only when the lockfile is uploaded by the current workflow (no matter whether the steps after uploading the lockfile succeed or fail). Failing to upload the lockfile means the swift container for the image is still occupied by other workflows.

src/image/validate_unique_trigger_upload_entry.py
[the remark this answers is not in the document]
zhijie-yang: Fixed in [076d2dc](https://github.com/canonical/oci-factory/pull/201/commits/076d2dc63c205a4489af33dde4b170e4d39c89fb)

.github/workflows/Image.yaml
```diff
@@ -105,17 +104,10 @@ jobs:
...
-          OS_TENANT_NAME: ${{ secrets.SWIFT_OS_TENANT_NAME }}
-          OS_PASSWORD: ${{ secrets.SWIFT_OS_PASSWORD }}
-          OS_REGION_NAME: ${{ secrets.SWIFT_OS_REGION_NAME }}
-          OS_STORAGE_URL: ${{ secrets.SWIFT_OS_STORAGE_URL }}
-          IMAGE_NAME: ${{ steps.validate-image.outputs.img-name }}
-          SWIFT_CONTAINER_NAME: ${{ vars.SWIFT_CONTAINER_NAME }}
-        run: ./src/image/define_image_revision.sh
+      - name: Validate unique trigger upload entry
+        run: |
+          ./src/image/validate_unique_trigger_upload_entry.py \
```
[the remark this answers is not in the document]
zhijie-yang: Fixed in [076d2dc](https://github.com/canonical/oci-factory/pull/201/commits/076d2dc63c205a4489af33dde4b170e4d39c89fb)

cjdcordeiro (approved):
ty for the changes, can we merge?

zhijie-yang:
Yes, I reverted the changes disabling ppc64el for mock-rock. We can merge now (even if the lpci build in the workflow can fail).

cjdcordeiro:
there's a conflict that needs resolution

zhijie-yang:
Done.

cjdcordeiro (commented):
why is the prometheus rock being built in the CI?

zhijie-yang:
Because of the merge commit [2acd80e](https://github.com/canonical/oci-factory/pull/201/commits/2acd80e0d72a790b8de30717513509e9b78d963a).

PR #203, opened 2024-06-21
Add 'deprecated' field for depreacted section generation in image docs

linostar:

---

## Description
The documentation generator for rocks is missing some parts necessary to generate the `Deprecated channels & tags` section.

This PR provides the necessary values so that the generator can identify the deperacted tags and list them accordingly, using the expired end-of-life value of each track to determine whether it is depreacted or not.

### Related issues
OCI-Factory builds won't work if one image's track has an expired EOL value. Another PR to fix that issue will be out soon.

### DIFF
--- a/src/docs/generate_oci_doc_yaml.py
+++ b/src/docs/generate_oci_doc_yaml.py
@@ -12,9 +12,8 @@ import re
 import subprocess
 import sys
 import tempfile
-import timezone
 from typing import Any, Dict, List
-from datetime import datetime
+from datetime import datetime, timezone
 from dateutil import parser
 
 import boto3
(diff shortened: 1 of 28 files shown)


src/docs/generate_oci_doc_yaml.py
```diff
@@ -269,6 +271,11 @@ def build_releases_data(
                     "until": eol.strftime("%m/%Y")
                 }
 
+                if eol > datetime.now(timezone.utc):
```
SamirPS, line 274: why don't use datetime.UTC? https://docs.python.org/3/library/datetime.html#datetime.UTC

cjdcordeiro (commented):
Re-triggered the tests

PR #205, opened 2024-07-03
Mattermost Notifier for Temporal Worker

zhijie-yang:

---

## Description
This commit adds a Mattermost notifier for the OCI Factory temporal worker. Due to the stability issue in the temporal Python SDK, the worker can be killed prematurely before the activity is done. Therefore, we need an MM notifier to notify the developers about the status (e.g., if the rebuild is triggered, and the triggering process succeeds or fails).

### Related issues

https://github.com/temporalio/sdk-python/issues/459


src/workflow-engine/charms/temporal-worker/oci_factory/activities/activity_consume_events.py
```diff
@@ -61,6 +74,11 @@ async def consume(topic: str, consumer_group: str) -> dict:
     proc = subprocess.Popen(
         ["python3", script_full_path, "{}".format(value.get("release"))]
     )
-    proc.wait()
+    ret_code = proc.wait()
+
+    status = "Success" if ret_code == 0 else "Failed"
```
SamirPS, line 79: For this, can it be a good idea to use the walrus operator to avoid  doing two times the if condition?

source: https://realpython.com/python-walrus-operator/

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
```diff
@@ -0,0 +1,100 @@
...
+        logging.warning("Unable to notify on Mattermost, MATTERMOST_TOKEN is not set!")
+
+    data = json.dumps(payload).encode("utf-8")
+    res = requests.post(posts_url, headers=headers, data=data, timeout=10)
+    res.raise_for_status()
+    return res.json()["id"]
+
+
+# Update the message conditionally based on the success of the operation
+def update_message_cond_success(
```
SamirPS, line 54: I would rename this one, you also update the message if the condition fail

src/workflow-engine/charms/temporal-worker/oci_factory/activities/activity_consume_events.py
zhijie-yang: Thanks for the suggestion. I found directly using `success = proc.wait() == 0` is more concise. Fixed in [2674da7](https://github.com/canonical/oci-factory/commit/2674da7918c9f2eeeb71b88e1e7bf506e31498ba).

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
zhijie-yang: Thanks, fixed in [2674da7](https://github.com/canonical/oci-factory/commit/2674da7918c9f2eeeb71b88e1e7bf506e31498ba).

src/workflow-engine/charms/temporal-worker/oci_factory/activities/activity_consume_events.py
```diff
@@ -3,11 +3,21 @@
...
+    send_message,
+    update_status_and_message,
+)
 
 import logging
 import os
 import subprocess
 
+MM_MESSAGE_TITLE = (
+    "[OCI Factory Temporal Workflow]: Image Rebuild for New Ubuntu Release"
```
cjdcordeiro, line 16: Can we print some useful information about the activity? today we only have this one but once we have more, it will be useful to trace where the message came from. we have `activity.info()` that provides a lot of metadata about the workflow and activity

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
```diff
@@ -0,0 +1,99 @@
...
+COLORS = {
+    "success": "#33CC33",
+    "failure": "#CC0000",
+    "unknown": "#B0B0B0",
+}
+
+channel_id = os.environ.get(
+    "MATTERMOST_CHANNEL_ID", "yax1c3pmm7natd5ebw3cngyrpw"
+)  # `ROCKS Bot Test` by default
+mattermost_server = "https://chat.canonical.com"
```
cjdcordeiro, line 18: can you pls make these secrets and fix the channel ID?

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
```diff
@@ -0,0 +1,99 @@
...
+    title = "[OCI Factory Temporal Worker] Image Rebuild on New Ubuntu Release"
+    id = send_message(title, message)
+    time.sleep(5)
+    update_status_and_message(
+        id, 0, message="**Release**: 100.04\n**Status**: Failed"
+    )
+    time.sleep(5)
+    update_status_and_message(
+        id, 1, message="**Release**: 100.04\n**Status**: Success"
+    )
```
cjdcordeiro, line 99: I guess this is a test ?

cjdcordeiro (changes requested):

src/workflow-engine/charms/temporal-worker/oci_factory/activities/activity_consume_events.py
zhijie-yang: Yes, we do have a bunch of other information from the `activity.Info` class. Which of them do you think is necessary to report in the MM notification? (FYI, [docs](https://python.temporal.io/temporalio.activity.Info.html).) I would say the only important info for now is the `[workflow_id](https://python.temporal.io/temporalio.activity.Info.html#workflow_id)` and `[workflow_run_id](https://python.temporal.io/temporalio.activity.Info.html#workflow_run_id)`.

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
zhijie-yang: If you think this is a better idea, we can make both secrets and define them in the `.env` file.

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
zhijie-yang: Yes, do you have any suggestions on where to move it? This module is not meant to be run as `__main__` by design.

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
zhijie-yang: Resolved in [a0dec2e](https://github.com/canonical/oci-factory/pull/205/commits/a0dec2edef902c5887971b9930fb7797fa581ade)

src/workflow-engine/charms/temporal-worker/oci_factory/activities/activity_consume_events.py
cjdcordeiro: Y the workflow id and run-id are useful. But also something about the actual activity running this - the activity-id. I'd need to see an example of an output of that object, to be sure of what the info actually looks like, but this is a good start

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
zhijie-yang: Fixed.

src/workflow-engine/charms/temporal-worker/oci_factory/activities/activity_consume_events.py
zhijie-yang: These three added.

src/workflow-engine/charms/temporal-worker/oci_factory/activities/activity_consume_events.py
```diff
@@ -52,6 +71,10 @@ async def consume(topic: str, consumer_group: str) -> dict:
             consumer.close()
 
     logging.info("Release: {}".format(value["release"]))
+    print("Release: {}".format(value["release"]))
```
SamirPS, line 74: why a print here ?

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
```diff
@@ -0,0 +1,93 @@
...
+
+COLORS = {
+    "success": "#33CC33",
+    "failure": "#CC0000",
+    "unknown": "#B0B0B0",
+}
+
+REQUIRED_ENVS = ["MATTERMOST_CHANNEL_ID", "MATTERMOST_TOKEN", "MATTERMOST_SERVER"]
+
+missing_envs = [env for env in REQUIRED_ENVS if os.environ.get(env) is None]
```
SamirPS, line 19: why don't use any function in python?

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
```diff
@@ -0,0 +1,93 @@
...
+                    "fallback": message,
+                    "title": title,
+                    "text": message,
+                    "color": COLORS["unknown"],
+                }
+            ]
+        },
+    }
+
+    token = os.environ.get("MATTERMOST_TOKEN", None)
```
SamirPS, line 54: why this? you have all in MATTERMOST_TOKEN

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
```diff
@@ -0,0 +1,93 @@
...
+    props["attachments"][0]["color"] = (
+        COLORS["success"] if success else COLORS["failure"]
+    )
+
+    if message:
+        props["attachments"][0]["text"] = message
+
+    payload = {"id": message_id, "props": props}
+
+    token = os.environ.get("MATTERMOST_TOKEN", None)
```
SamirPS, line 85: MATTERMOST_TOKEN

src/workflow-engine/charms/temporal-worker/oci_factory/activities/activity_consume_events.py
zhijie-yang: To have the temporal sdk catch the stdout.

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
zhijie-yang: Good catch. Removed.

src/workflow-engine/charms/temporal-worker/oci_factory/notification/mattermost_notifier.py
SamirPS: https://www.programiz.com/python-programming/methods/built-in/any

SamirPS: with a lambda

zhijie-yang: Using the list comprehension here makes it possible to print an error message informing the user all the missing environmental variables. I don't see a possibility using `any` could achieve so without over-complicating the code.

src/workflow-engine/charms/temporal-worker/oci_factory/activities/activity_consume_events.py
```diff
@@ -3,11 +3,27 @@
...
+    update_status_and_message,
+)
 
 import logging
 import os
 import subprocess
 
+
+TWC_HOST = os.environ.get("TWC_HOST")
+TWC_NAMESPACE = os.environ.get("TWC_NAMESPACE")
```
cjdcordeiro, line 17: oh I didn't know these were available! that's good

cjdcordeiro (changes requested):
could you pls fix some of the simple pylinting errors like long lines , import order, docstrings, etc.

cjdcordeiro (approved):
better now, thanks. There are still error, but I think those are due to import errors that we can fix later during the refactoring

PR #210, opened 2024-07-23
feat: add python3.11

zhijie-yang:

---

## Description
This PR adds Python 3.11 to the OCI Factory.

### DIFF
--- a/oci/python/image.yaml
+++ b/oci/python/image.yaml
@@ -2,7 +2,7 @@ version: 1
 
 upload:
   - source: canonical/chiselled-python
-    commit: 26d72f441aa2ad3bcd830cee6d6996b384be223c
+    commit: ab1bf03d14a627e6153045cb5d5f65870ca02fc4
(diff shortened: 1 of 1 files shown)


linostar (approved):
LGTM, thanks!

cjdcordeiro (changes requested):
we do **not** want to publish 3.11. that would be a Pro rock.

cjdcordeiro:
we still need to rebuild 3.8 and 3.12. Have there been no changes to those since the last oci factory PR?

zhijie-yang:
There are. Besides the change for python 3.10, there is only a minor change: The chisel version used in the Dockerfile of python 3.8 is been bumped from 0.9.0 to 0.10.0. Since the python images are all build by the oci factory with rockcraft, this should make no difference.

cjdcordeiro:
ok then we still need your 3.10 change and I'll manually re-trigger 3.8

PR #212, opened 2024-07-24
fix: move upload trigger condition to prepare-upload

zhijie-yang:

---

## Description
This PR fix the issue that the `prepare-upload` job is not protected by the triggering condition, causing unnecessary increments of revision numbers if the workflow is not expected to run the `upload` job.

### DIFF
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -149,6 +149,7 @@ jobs:
     runs-on: ubuntu-22.04
     needs: [prepare-build, run-build, test]
     name: Prepare upload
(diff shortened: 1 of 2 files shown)


cjdcordeiro (changes requested):
that's a new one: https://github.com/canonical/oci-factory/actions/runs/10072927752/job/27848332935?pr=212

zhijie-yang:
The problem seems to be from Launchpad git side. Retry on the failed jobs now succeeds.

cjdcordeiro (approved):
great thanks

PR #213, opened 2024-07-25
fix(ci): pass commit sha to force doc wf checkout to newest main

zhijie-yang:

---

## Description

This PR fixes the issue when a new track is released with the OCI Factory, the EOL field is not properly set after the first run of the `Image.yaml` workflow. The cause of the problem is the default checkout refs in the called workflow `Documentation.yaml` is the HEAD of main when the `Image.yaml` workflow is dispatched, in which the updates to `_releases.json` in the `Release.yaml` workflow is not included yet. The PR passes the commit SHA that contains the updates to `_releases.json` from the `Release.yaml` workflow, and forces the `Documentation.yaml` workflow to checkout to this commit when generating the document yaml.

### Temporary solution for the issue

Manually trigger the `Documentation.yaml` workflow one more time after the `Image.yaml` workflow completes with success. 

### Related issues
ROCKS-1354
The [checked commit in Documentation](https://github.com/canonical/oci-factory/actions/runs/10080015719/job/27869397538#step:2:69) is the one prior to the [commit contains _releases.json updates](https://github.com/canonical/oci-factory/actions/runs/10080015719/job/27869295243#step:10:34)

### DIFF
--- a/.github/workflows/Documentation.yaml
+++ b/.github/workflows/Documentation.yaml
@@ -83,7 +83,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          ref: ${{ inputs.release-commit-sha}}
+          ref: ${{ inputs.release-commit-sha }}
       
       - uses: actions/setup-python@v5
         with:
--- a/.github/workflows/Documentation.yaml
+++ b/.github/workflows/Documentation.yaml
@@ -21,6 +21,10 @@ on:
         description: 'OCI image to generate the documentation for'
         required: true
         type: string
+      release-commit-sha:
+        description: 'Commit SHA containing the updated _releases.json'
+        required: true
+        type: string
 
 jobs:
   validate-documentation-request:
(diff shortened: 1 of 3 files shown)


.github/workflows/Documentation.yaml
```diff
@@ -78,6 +82,8 @@ jobs:
       IS_PROD: ${{ ! startsWith(needs.validate-documentation-request.outputs.oci-img-name, 'mock-') }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.release-commit-sha}}
```
cjdcordeiro, line 86:
```suggestion
          ref: ${{ inputs.release-commit-sha }}
```

cjdcordeiro: and, what happen on `workflow_dispatch`? Will this be empty or throw an error? and if empty, will the `checkout` action default to `main`?

cjdcordeiro (changes requested):

.github/workflows/Documentation.yaml
zhijie-yang: On `workflow_dispatch` the `inputs.release-commit-sha` will be empty. The `with: ref:` won't be effective. The checkout action will default to the branch, based on which the workflow is dispatched.

cjdcordeiro (approved):

linostar (approved):
Looks good, thanks for the fix.

PR #217, opened 2024-08-07
Add Access Control to Dispatchable Workflows

zhijie-yang:

---

## Description
This PR adds the access control to the dispatchable workflow, in order to have the right user access to workflow when external users are granted "write" access.

### Related issues
Not Applicable.

## Tests
1. With @zhijie-yang in CODEOWNER: Should pass: [run1](https://github.com/canonical/oci-factory/actions/runs/10301256506) and [run2](https://github.com/canonical/oci-factory/actions/runs/10301820520)
2. With @zhijie-yang neither in CODEOWNER nor in `mock-rock/contacts.yaml` as maintainer: Should cancel: [run](https://github.com/canonical/oci-factory/actions/runs/10828699927/job/30044653419#step:3:31)
3. With @zhijie-yang in `mock-rock/contacts.yaml` as maintainer: Should pass: [run](https://github.com/canonical/oci-factory/actions/runs/10301739295/job/28513942445#step:3:41)
4. Reusing the public workflows for other repos skips access validation: [run](https://github.com/zhijie-yang/oci-factory/actions/runs/10507113064/job/29108176625)
---
Latest workflow run: https://github.com/canonical/oci-factory/actions/runs/10828858187


.github/actions/validate-access/action.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+runs:
+  using: "composite" 
+  steps:
+    - name: Check if the workflow is triggered by an admin user
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
+      run: |
+        echo "github.actor: ${{ github.actor }}"
```
clay-lake, line 29: I think we should break this out into a script and add some testing to ensure that it fails safe. I'm currently adding testing elsewhere with pytest, maybe we can integrate that somehow. Otherwise there is also https://github.com/sstephenson/bats

We could also start the script `set -e` for added safety.

.github/actions/validate-access/action.yaml
```diff
@@ -0,0 +1,53 @@
...
+        echo "admin-only: ${{ inputs.admin-only }}"
+        if [[ ${{ inputs.admin-only }} == true ]]; then
+          echo "Expanding team mentions in the CODEOWNERS file"
+          teams=$(grep -oE '@[[:alnum:]_.-]+\/[[:alnum:]_.-]+' ${{ github.workspace }}/CODEOWNERS || true | sort | uniq)
+
+          for team in $teams; do
+            org=$(echo $team | cut -d'/' -f1 | sed 's/@//')
+            team_name=$(echo $team | cut -d'/' -f2)
```
clay-lake, line 37: I'm not sure about the name restrictions on $team, but maybe we should double quote it to be safe?

.github/actions/validate-access/action.yaml
```diff
@@ -0,0 +1,53 @@
...
+            members=$(gh api "/orgs/$org/teams/$team_name/members" | jq -r '.[].login')
+            replacement=$(echo "$members" | xargs -I {} echo -n "@{} " | awk '{$1=$1};1')
+            sed -i "s|$team|$replacement|g" ${{ github.workspace }}/CODEOWNERS
+          done
+
+          if grep -wq "@${{ github.actor }}" ${{ github.workspace }}/CODEOWNERS; then
+            echo "The workflow is triggered by ${{ github.actor }} as the code owner"
+          elif cat ${{ github.workspace }}/${{ inputs.image-path }}/contacts.yaml | yq ".maintainers" | grep "\- " | grep -wq "${{ github.actor }}"; then
```
clay-lake, line 45:
```suggestion
          elif cat "${{ github.workspace }}/${{ inputs.image-path }}/contacts.yaml" | yq ".maintainers" | grep "\- " | grep -wq "${{ github.actor }}"; then
```

clay-lake (changes requested):

.github/actions/validate-access/action.yaml
zhijie-yang: Breaking it into a standalone script is a good idea. Setting `-e` is not necessary, as running the composite action with `bash` is actually `bash --noprofile --norc -e -o pipefail`, which already sets this.

.github/actions/validate-access/action.yaml
zhijie-yang: IMO, we don't have to, since the $team does not contain white spaces.

.github/actions/validate-access/action.yaml
zhijie-yang: The `${{ github.workspace }}` and `${{ inputs.image-path }}` is not the shell syntax, but GitHub Workflow expression syntax instead. Since we are not passing any white spaces in these variables, we don't have to quote them.

.github/actions/validate-access/test-validate-access.bats
cjdcordeiro, line 1: This is nice! Thanks.
It must be run as part of the CI though, otherwise the tests are never checked on changes

.github/workflows/Documentation.yaml
```diff
@@ -25,11 +27,16 @@ on:
         description: 'Commit SHA containing the updated _releases.json'
         required: true
         type: string
+      is-workflow_call:
```
cjdcordeiro, line 30: you don't need this. just use `${{ github.event_name }}`

.github/actions/validate-access/action.yaml
```diff
@@ -0,0 +1,28 @@
+name: Validate Access
```
cjdcordeiro, line 1:
```suggestion
name: Validate Actor
```

Can you change the remaining names accordingly pls? This is more descriptive of the action cause that's what it does - check if the `actor` has the right permissions

.github/workflows/Documentation.yaml
```diff
@@ -25,11 +27,16 @@ on:
...
+        description: 'Flag to indicate if this is a workflow call'
+        type: boolean
+        default: true
 
 jobs:
   validate-documentation-request:
     runs-on: ubuntu-22.04
     name: Validate documentation request
+    # needs: [access-check]
```
cjdcordeiro, line 39: leftover?

.github/actions/validate-access/action.yaml
```diff
@@ -0,0 +1,28 @@
...
+    description: 'The GH_TOKEN for the GitHub CLI'
+    required: true
+
+runs:
+  using: "composite" 
+  steps:
+    - name: Check if the workflow is triggered by an admin user
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
```
cjdcordeiro, line 27:
```suggestion
        GITHUB_TOKEN: ${{ inputs.github-token }}
```

for consistency

cjdcordeiro: I think it should also work with `gh`, but if it doesn't feel free to reject this suggestion

.github/workflows/Tests.yaml
```diff
@@ -66,9 +70,23 @@ env:
...
+    steps:
+    - uses: actions/checkout@v4
+      if: ${{ !inputs.is-workflow_call }}
+    - name: Validate access to triggered image
+      uses: ./.github/actions/validate-access
+      if: ${{ !inputs.is-workflow_call }}
+      with:
+        admin-only: true
+        image-path: ${{ inputs.oci-image-path }}
+        github-token: ${{ secrets.ROCKSBOT_TOKEN }}
```
cjdcordeiro, line 84: doesn't the GITHUB_TOKEN work?

cjdcordeiro (changes requested):

.github/workflows/Tests.yaml
zhijie-yang: No, since the `team` attribute of an organization is private to internal uses only.

.github/actions/validate-access/test-validate-access.bats
zhijie-yang: Yeah, but I think it's better to wait for @clay-lake 's changes on the unit tests on CI scripts.

.github/actions/validate-access/action.yaml
zhijie-yang: Fixed.

.github/workflows/Documentation.yaml
zhijie-yang: Fixed.

.github/actions/validate-access/action.yaml
zhijie-yang: It amazingly works! Fixed.

.github/workflows/Documentation.yaml
cjdcordeiro: Yes that's true because it will depend on where it's being called.

But that highlights exactly the issue you have here:
 - sometimes, even if being used as a workflow_call, you may want to run the validation check (e.g. the `Documentation` workflow is called from a remotely dispatched workflow) (in which case it is good that `event_name` is `workflow_dispatch`

I know - "the dispatched workflow already does the validation". But that's not future proof and creates a loose dependency that can very easily break.

We need a more robust condition to decide when to make the check.
I'm thinking about using a combo of `github.repository` (we only care about validation when running in the `oci-factory` repo - others who _call this from outside don't need this validation) with `github.event` (the name is still useful, i.e. if it says `workflow_dispatch`, then you know you want to run it, even if this workflow was called by another) with `inputs` (your approach is good but introduces an unnecessary variable - just use another that is unique to the workflow_call)

cjdcordeiro (commented):
(just leaving a comment so I get pinged again when this is ready for review)

clay-lake (approved):
Nice! Thanks for update to the Cancel behavior.

PR #224, opened 2024-08-21
Feat: add pytest job to _Test-OCI-Factory workflow

clay-lake:

This PR adds a job to `.github/workflows/_Test-OCI-Factory.yaml` which executes pytest from the root of the repository and generates a simple test report to that step's `$GITHUB_STEP_SUMMARY`.

More tests will be added in subsequent refactoring PRs


clay-lake:
Here is a sample integration where you can see the test summary, and the uploaded jUnit report artifact. 
https://github.com/clay-lake/oci-factory/actions/runs/10473012228

.github/workflows/_Test-OCI-Factory.yaml
```diff
@@ -15,7 +15,58 @@ on:
       - "!src/workflow-engine/**"
       - "!src/cli-client/**"
 
+env:
+  # local path to clone the oci-factory to
+  OCI_FACTORY_DIR: oci-factory/
```
cjdcordeiro, line 20:
```suggestion
```

it's unnecessary - the checkout action will provide the repo at `PWD`

.github/workflows/_Test-OCI-Factory.yaml
```diff
@@ -15,7 +15,58 @@ on:
...
+  pytest:
+    # Trigger python unit tests across the repository
+    name: pytest
+    runs-on: ubuntu-22.04
+    steps:
+
+      # Job Setup
+      - uses: actions/checkout@v4
+        with:
+          path: ${{ env.OCI_FACTORY_DIR }}
```
cjdcordeiro, line 37:
```suggestion
```

.github/workflows/_Test-OCI-Factory.yaml
```diff
@@ -15,7 +15,58 @@ on:
...
+          fetch-depth: 1
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
+      
+      # Note: Add additional dependency installation lines as required below 
+      # test-oci-factory/pytest requirements
+      - run: pip install -r ${{ env.OCI_FACTORY_DIR }}/src/test-oci-factory/pytest/requirements.txt
```
cjdcordeiro, line 46:
```suggestion
      - run: pip install -r src/test-oci-factory/pytest/requirements.txt
```

(same for other occurrences below)

.github/workflows/_Test-OCI-Factory.yaml
```diff
@@ -15,7 +15,58 @@ on:
...
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ env.PYTEST_RESULT_PATH }}
+          path: ${{ env.PYTEST_RESULT_PATH }}
+          if-no-files-found: error
+
+
+
+
```
cjdcordeiro, line 69:
```suggestion


```

src/test-oci-factory/pytest/format_markdown_report.py
cjdcordeiro, line 1: the `test-oci-factory/pytest` folder name is a bit awkward. 

I'd suggest following a more python project-like structure, moving this folder and all subsequent unit test into the root path, at `tests/`, such that you can then have `tests/{unit/,data/,requirements.txt,utils/...}`

cjdcordeiro (changes requested):
This sounds like a good initiative. 

I'd like to see this combined with at least one test, so I'd wait at least until 1 test is ready.

.github/workflows/_Test-OCI-Factory.yaml
clay-lake: I started doing this by convention because later in the build workflow we clone 2 repositories, the rock-repo for the project, and our own repo for the scripts. It is little awkward, but it prevents collisions if the repos had similar directory structures.

cjdcordeiro: that won't happen cause those are different jobs. Each job has it own workspace, so there won't be any collisions

clay-lake: Here is a sample from the refactored build-rock workflow where it could happen in the lpci job. If we get rid of this type of build, we could also mitigate the collision issue entirely. What do you think?


```
      # Job Setup
      - uses: actions/checkout@v4
        with:
          repository: ${{ env.OCI_FACTORY_REPO }}
          path: ${{ env.OCI_FACTORY_DIR }}
          ref: ${{ env.OCI_FACTORY_BRANCH }}
          fetch-depth: 1

      - uses: actions/setup-python@v5
        with:
          python-version: "3.x"
      - run: pip install -r ${{ env.OCI_FACTORY_DIR }}/src/build-rock/lpci_build/requirements.txt
      

      - name: Clone Repo
        uses: actions/checkout@v4
        with:
          repository: ${{ inputs.rock-repo }}
          path: ${{ env.ROCK_REPO_DIR }}
          ref: ${{ inputs.rock-repo-commit }}
          submodules: 'recursive'

      - name: Build Target
      # TODO: Replace this retry action with bash equivalent for better testing
        uses: nick-fields/retry@v3.0.0

...
``

File: src/test-oci-factory/pytest/format_markdown_report.py
Comment (clay-lake) [reply src/test-oci-factory/pytest/format_markdown_report.py:1]: The `test-oci-factory/pytest` directory contains code that supports the `pytest` job in `_Test-OCI-Factory.yaml`. So far, the code in this directory only creates a report for redirection to `$GITHUB_STEP_SUMMARY`. I'm not planning on it being _the_ test directory.

I plan to add further tests in each module of the repository, so the code remains relatively compartmentalized. For example the jobs` assemble-rock`, `configure` and `lpci_build` of the build-rock workflow each have a directory. 

The scripts in the `configure` job have a few tests in the script prefixed with 'test_'

```
build-rock
├── assemble-rock
│  ├── assemble.sh
│  └── requirements.sh
├── configure
│  ├── generate_build_matrix.py
│  ├── requirements.txt
│  └── test_generate_build_matrix.py
└── lpci_build
   └── lpci_build.sh

```
Executing pytest on the repo root will collect the tests and individually and execute them per module, and the summary looks like this: https://github.com/clay-lake/oci-factory/actions/runs/10473012228

File: .github/workflows/_Test-OCI-Factory.yaml (hunk 1 above)
Comment (cjdcordeiro) [reply .github/workflows/_Test-OCI-Factory.yaml:20]: But that's a different job. In that case, it makes sense. But in this job it is unnecessary.

Comment (clay-lake) [reply .github/workflows/_Test-OCI-Factory.yaml:20]: Sounds good, I will apply the changes.

File: src/test-oci-factory/pytest/format_markdown_report.py
Comment (cjdcordeiro) [reply src/test-oci-factory/pytest/format_markdown_report.py:1]: That's also a good structure.

This is the kind of structure I was thinking about: https://github.com/canonical/rockcraft/tree/main/tests/unit

I don't have a strong preference TBH. Python projects seem to commonly do `src` and `test` at root level (eg. Rockcraft , or https://github.com/pytest-dev/pytest), while Golang projects seem to adopt something closer to your proposal.

The only thing hitching me is this additional "test-oci-factory" folder, inside `src`, when it isn't actually a `src`.

File: .github/workflows/_Test-OCI-Factory.yaml [hunk 5]
```diff
@@ -15,7 +15,58 @@ on:
...
+      - run: pip install -r ${{ env.OCI_FACTORY_DIR }}/src/test-oci-factory/pytest/requirements.txt
+
+
+      - name: Run pytest
+        continue-on-error: true
+        run: |
+          python3 -m pytest --junit-xml "${{ env.PYTEST_RESULT_PATH }}" "${{ env.OCI_FACTORY_DIR }}"
+
+      - name: Generate Summary
+        if: always()
```
zhijie-yang, line 55: We can use ${{ !cancelled() }} instead since the workflow can be interrupted at previous steps --- A suggestion from the [Github Documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always).

.github/workflows/_Test-OCI-Factory.yaml
```diff
@@ -15,7 +15,58 @@ on:
...
+        run: |
+          python3 ${{ env.OCI_FACTORY_DIR }}/src/test-oci-factory/pytest/format_markdown_report.py --input-junit "${{ env.PYTEST_RESULT_PATH }}" >> $GITHUB_STEP_SUMMARY
+
+      - name: Upload pytest Result
+        if: always()
```
zhijie-yang, line 60:
```suggestion
        if: ${{ !cancelled() }}
```

zhijie-yang (changes requested):

src/test-oci-factory/pytest/format_markdown_report.py
clay-lake: The rest of our org is placing tests outside of source. Lets do that so its's regular with our other repos. That solves your final point as well.

.github/workflows/_Test-OCI-Factory.yaml
clay-lake: Nice, lets do that

linostar (approved):
Looks neat and good. My remarks have already been addressed.

.github/workflows/_Test-OCI-Factory.yaml
cjdcordeiro, line 1: You need to change the `paths` such that this workflow also runs now with the new `tests` and `tools`. You might want to skip the `test-workflows:` for those paths.

Needed by https://github.com/canonical/oci-factory/pull/217#discussion_r1723580592

cjdcordeiro (changes requested):
almost there. You may also want to lint all files. Some are missing a newline at EOF and others have inconsistent break lines throughout the text

.github/workflows/_Test-OCI-Factory.yaml
clay-lake: Thanks! I added the two paths. Let's see how it runs without excluding the test-workflows job. Hopefully we can keep keep the tests locally executable (or at least mark the GH runner only ones) so we don't need to push for testing.

cjdcordeiro: there's a conflict it seems

zhijie-yang (approved):
Looks good to me now. Thanks!

cjdcordeiro (approved):
greeeat thanks

PR #233, opened 2024-08-23
ROCKS 1452 - Refactor Build Rock workflow to be externally reusable

clay-lake:

_Note: This PR depends on #224_ 
This pull request refactors the build workflow to enable re-usability by external repositories. There is an example application of this in the rocks-toolbox. A example job run can be found here: https://github.com/canonical/rocks-toolbox/actions/runs/10524164671


clay-lake:
Since it is not part of the checks, here is an example Test OCI Factory run from e6e3c29
https://github.com/canonical/oci-factory/actions/runs/10678234515

.github/workflows/Image.yaml
```diff
@@ -115,23 +115,47 @@ jobs:
...
     secrets: inherit
 
-  test:
+  tmp-cache-job:
+    # Note: This is a temporary job that will be removed when the refactored test job is merged.
+    # Going forward we download the built rocks from artifacts instead of cache. This job takes
+    # the uploaded rocks then re-caches them for compatibility. 
+    name: Temporary step to cache rocks
+    runs-on: ubuntu-22.04
     needs: [prepare-build, run-build]
```
zhijie-yang, line 140:
```suggestion
    needs: [run-build]
```

.github/workflows/Image.yaml
```diff
@@ -115,23 +115,47 @@ jobs:
...
+          name: ${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.dir_identifier }}
+
+      - uses: actions/cache/save@v4
+        with:
+          key: ${{ github.run_id }}-${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.dir_identifier }}  
+          path: ${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.dir_identifier }}
+
+
+  test:
+    needs: [prepare-build, run-build, tmp-cache-job]
```
zhijie-yang, line 158:
```suggestion
    # TODO: Remove tmp-cache-job when removing the job tmp-cache-job
    needs: [run-build, tmp-cache-job]
```

src/build_rock/configure/generate_build_matrix.py
```diff
@@ -0,0 +1,139 @@
...
+        github_output.write(fh, **outputs)
+
+
+def main():
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "--rockfile-directory",
+        help="Path where to directory containing rockcraft.yaml.",
+        required=True,
```
zhijie-yang, line 95:
```suggestion
        required=True,
        type=str
```

src/build_rock/lpci_build/lpci_build.sh
```diff
@@ -0,0 +1,58 @@
...
+}
+
+while getopts "c:d:" opt
+do
+    case $opt in
+        d)
+            ROCKCRAFT_DIR="$OPTARG"
+            ;;
+        c)
+            LC_CREDENTIALS_B64="$OPTARG"
```
zhijie-yang, line 24:
```suggestion
            LP_CREDENTIALS_B64="$OPTARG"
```
Also for the following occurances.

src/build_rock/configure/generate_build_matrix.py
```diff
@@ -0,0 +1,139 @@
...
+    """Update GITHUB_OUTPUT with build configuration."""
+
+    outputs = {"rock-name": rock_name, **build_matrices}
+
+    # set default when not testing
+    if output_path is None:
+        output_path = os.environ["GITHUB_OUTPUT"]
+
+    with open(output_path, "a") as fh:
+        github_output.write(fh, **outputs)
```
zhijie-yang, line 86: Do we need to manually set the `output_path`? In GH runners, `os.environ["GITHUB_OUTPUT"]` is already set to some file. For testing, we can just inject the `GITHUB_OUTPUT` environmental variable locally.

src/shared/github_output.py
zhijie-yang, line 1: This file abstracts the writes to the file `GITHUB_OUTPUT`, so we should use this module to determine where is the destination we want to write the outputs. As mentioned above, `GITHUB_OUTPUT` is an environmental variable designated in the GH workflow to handle the outputs. For the test purpose, we can just inject the `GITHUB_OUTPUT` with an arbitrary file path or even `/dev/stdout`, since we don't need the `GITHUB_OUTPUT` to be mutable during the run. With this, we can avoid letting the callers decide where to output, and provide a better abstracted interface to the callers.

tests/fixtures/__ini__.py
zhijie-yang, line 1: I assume the file name contains a typo. It should be `__init__.py` instead.

tests/integration/test_convert_junit_xml_to_markdown.py
```diff
@@ -0,0 +1,18 @@
+from ..fixtures.buffers import str_buff
+from ..fixtures.junit_et import junit_with_failure
```
zhijie-yang, line 2: We are missing the fixture files in the `fixtures` directory.

tests/unit/test_generate_build_matrix.py
```diff
@@ -0,0 +1,101 @@
...
+                "environment": {"VAR1": "value", "VAR2": "other value"},
+            }
+        },
+        "platforms": {
+            "amd64": None,
+            "armhf": {"build-for": ["armhf", "arm64"]},
+            "ibm": {"build-on": ["s390x"], "build-for": "s390x"},
+        },
+        "parts": {"hello": {"plugin": "nil", "stage-packages": ["hello"]}},
+    }
```
zhijie-yang, line 40: We can use a fixture to store this constant string ;D

zhijie-yang (changes requested):
Thanks for the changes! Leaving several nitpicks:

tests/unit/test_github_output.py
```diff
@@ -0,0 +1,54 @@
...
+from src.shared.github_output import write, format_value
+from io import StringIO
+import pytest
+
+
+@pytest.fixture
+def text_buffer():
+    with StringIO() as buffer:
+        yield buffer
+
```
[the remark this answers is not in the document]
clay-lake: Nice, thanks for catching that.

tests/integration/test_convert_junit_xml_to_markdown.py
clay-lake: Could you send me a screenshot? (MM works). Git is tracking these files:

```
:~/source/oci-factory-prs$ git log tests/fixtures/buffers.py tests/fixtures/junit_et.py
commit ca94d3ad4b2ff765d0100ad35ab373d51a97f141
Author: Adrian Clay Lake <clay.lake@canonical.com>
Date:   Thu Aug 22 12:09:26 2024 +0200

    feat: added tests directory and refactored accordingly
```

tests/fixtures/__ini__.py
clay-lake: yes! whoops

.github/workflows/Image.yaml
```diff
@@ -115,23 +115,47 @@ jobs:
             --oci-path ${{ steps.validate-image.outputs.img-path }} \
             --revision-data-dir ${{ env.DATA_DIR }} \
 
+
```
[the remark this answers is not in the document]
clay-lake: Done!

.github/workflows/Build-Rock.yaml
```diff
@@ -1,206 +1,227 @@
...
+          if-no-files-found: error
+
+
+  lpci-build:
+    # lpci-build is a fallback for building rocks if no suitable runners are 
+    # configured for the required architecture. Builds in this job will be
+    # outsourced to Launchpad for completion. 
+    # Note the Secret
+    needs: [configure-build]
+    if: fromJSON(needs.configure-build.outputs.lpci-build-matrix).include[0] != ''
```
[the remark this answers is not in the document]
clay-lake: Let me see if I can find an old build run. The runner-build uses the same conditional, and it triggers as proof.

src/build_rock/configure/generate_build_matrix.py
```diff
@@ -0,0 +1,139 @@
...
+        help="Path where to directory containing rockcraft.yaml.",
+        required=True,
+    )
+
+    parser.add_argument(
+        "--lpci-fallback",
+        help="Revert to lpci if architectures are not supported. <true|false>",
+        required=True,
+        type=str,
+    )
```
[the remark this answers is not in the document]
clay-lake: Done. Thanks!

.github/workflows/Build-Rock.yaml
```diff
@@ -1,206 +1,227 @@
...
-        description: "Directory, in 'rock-repo', where to find the rockcraft.yaml file"
+        description: "Directory in repository where to find the rockcraft.yaml file."
         type: string
         required: true
 
+
 env:
-  ROCKS_CI_FOLDER: ci-rocks
+  ROCK_REPO_DIR: rock-repo  # path where the image repo is cloned to
+  ROCK_CI_FOLDER: ci-rocks  # path of uploaded/downloaded artifacts
```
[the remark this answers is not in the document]
clay-lake: That is fair. If it isn't a necessity I would keep this until my next PR as it uses the same language. Then we can change them all in one go.

.github/workflows/Build-Rock.yaml
```diff
@@ -1,206 +1,227 @@
...
         uses: actions/upload-artifact@v4
         with:
           name: ${{ inputs.oci-archive-name }}
           path: ${{ inputs.oci-archive-name }}
-          if-no-files-found: error
-      - uses: actions/cache/save@v4
-        with:
-          path: ${{ inputs.oci-archive-name }}
-          key: ${{ github.run_id }}-${{ inputs.oci-archive-name }}
+          if-no-files-found: error
```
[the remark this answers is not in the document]
clay-lake: Done. Thanks!

src/build_rock/configure/generate_build_matrix.py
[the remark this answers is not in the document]
clay-lake: I like the idea of breaking it out into a function to cast the type. Maybe it would be easier to match the string directly, that way I can still have somewhat flexible inputs.

.github/workflows/Image.yaml
```diff
@@ -115,23 +115,47 @@ jobs:
...
       rockfile-directory: ${{ matrix.directory }}
+      arch-map: '{"amd64": "ubuntu-22.04", "arm64": "Ubuntu_ARM64_4C_16G_01"}'
+      lpci-fallback: true
     secrets: inherit
 
-  test:
+  tmp-cache-job:
+    # Note: This is a temporary job that will be removed when the refactored test job is merged.
+    # Going forward we download the built rocks from artifacts instead of cache. This job takes
+    # the uploaded rocks then re-caches them for compatibility.
```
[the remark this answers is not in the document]
clay-lake: Nice! Done

.github/workflows/Build-Rock.yaml
```diff
@@ -1,206 +1,227 @@
...
 
+
   assemble-rock:
-    needs: [prepare-multi-arch-matrix, build]
+    # Assemble individual single-arch rocks into multi-arch rocks
+    needs: [configure-build, runner-build, lpci-build]
     runs-on: ubuntu-22.04
+    # Always run even if one of the *-build jobs are skipped
+    # Nice example from benjamin-bergia/github-workflow-patterns...
+    if: ${{ always() && contains(needs.*.result, 'success') && !(contains(needs.*.result, 'failure')) }}
```
linostar, line 195: this can be simplified to just `if: ${{ always() && !contains(needs.*.result, 'failure') }}`

src/build_rock/configure/generate_build_matrix.py
```diff
@@ -0,0 +1,139 @@
...
+def get_target_archs(rockcraft: dict) -> list:
+    """get list of target architectures from rockcraft project definition"""
+
+    rock_platforms = rockcraft["platforms"]
+
+    target_archs = set()
+
+    for platf, values in rock_platforms.items():
+
+        if isinstance(values, dict) and "build-for" in values:
```
linostar, line 30: I believe "build-for" is correct here.

src/build_rock/configure/generate_build_matrix.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # raise exception if we cannot fallback to LP builds
+        if not lp_fallback:
+            raise MissingArchSupport(
+                f"Missing support for runner arches: {missing_archs}"
+            )
+
+        # configure LP build
+        build_matrices[MATRIX_NAMES.LPCI.value]["include"].append(
+            {"architecture": "-".join(set(target_archs))}
```
linostar, line 60: you would want to build only `missing_archs` with lpci, not all of them, right?

linostar (changes requested):
Good work. I left few minor remarks and nitpicks.

.github/workflows/Image.yaml
clay-lake: This job is required for
```
matrix: ${{ fromJSON(needs.prepare-build.outputs.build-matrix) }
```

.github/workflows/Image.yaml
clay-lake: Thanks! Added

src/build_rock/configure/generate_build_matrix.py
clay-lake: Added! Thanks

src/build_rock/lpci_build/lpci_build.sh
clay-lake: Whoops! Thanks

tests/integration/test_convert_junit_xml_to_markdown.py
clay-lake: Re index files from dangling commits. You should see them now.

src/build_rock/configure/generate_build_matrix.py
clay-lake: That is a good question! I saw this earlier and I left this for later optimization. AFAIK this is the original behavior if you checkout the ppc64el-amd64 build logs here: https://github.com/canonical/oci-factory/actions/runs/10072927752/job/27849134914#logs

Do we need to build on the runner or lcpi exclusively?

linostar: Actually, lpci currently builds for amd64 and ppc64le, although amd64 is already being built with github runners. I believe Cris did that so that in the case where the ppc64le build fails we can compare it the lpci's amd64 build and see what went wrong. Github runners only support amd64 and arm64, so lpci was used to build extra architectures, but it is going to be replaced soon by launchpad builds.

.github/workflows/Image.yaml
```diff
@@ -115,23 +115,47 @@ jobs:
...
     uses: ./.github/workflows/Build-Rock.yaml
     with:
-      oci-archive-name: ${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.dir_identifier }}
-      oci-factory-path: ${{ matrix.path }}
-      rock-name: ${{ matrix.name }}
+      oci-archive-name: ${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.dir_identifier }}        
       rock-repo: ${{ matrix.source }}
       rock-repo-commit: ${{ matrix.commit }}
       rockfile-directory: ${{ matrix.directory }}
+      arch-map: '{"amd64": "ubuntu-22.04", "arm64": "Ubuntu_ARM64_4C_16G_01"}'
```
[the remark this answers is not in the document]
clay-lake: Works well! Updated.

src/build_rock/configure/generate_build_matrix.py
clay-lake: That is a good suggestion. I migrated these functions to their own class w/ context manager for simplicity.

src/shared/github_output.py
clay-lake: This should be handled as well as per above comment.

tests/unit/test_generate_build_matrix.py
clay-lake: Done. We now have a fixture for this at tests/fixtures/sample_data.py

src/build_rock/configure/generate_build_matrix.py
clay-lake: Thanks :smiley: Do you know of any projects that currently support this cross compilation feature?

.github/workflows/Build-Rock.yaml
clay-lake: We should have at least one `success`  for packaging, or else we wont have anything to operate on. In the current configuration we use either runner builder or the lpci builder.

.github/workflows/Build-Rock.yaml
```diff
@@ -1,206 +1,227 @@
...
-          path: ${{ env.ROCKS_CI_FOLDER }}
+          name: ${{ inputs.oci-archive-name }}-${{ steps.collect-artifacts.outputs.filename }}
+          path: ${{ env.ROCK_CI_FOLDER }}
           if-no-files-found: error
 
+
   assemble-rock:
-    needs: [prepare-multi-arch-matrix, build]
+    # Assemble individual single-arch rocks into multi-arch rocks
+    needs: [configure-build, runner-build, lpci-build]
```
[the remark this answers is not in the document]
clay-lake: Good catch!. Done

.github/workflows/Build-Rock.yaml
```diff
@@ -1,206 +1,227 @@
...
+
+
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v4
         with:
-          python-version: '3.x'
-      - uses: nick-fields/retry@v3.0.0
-        name: Build multi-arch ${{ inputs.rock-name }} in Launchpad
-        if: needs.prepare-multi-arch-matrix.outputs.build-with-lpci == 1
+          name: ${{ inputs.oci-archive-name }}-${{ steps.collect-artifacts.outputs.filename }}
```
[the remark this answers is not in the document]
clay-lake: We use that for adding a namespace to the builds, but! I just checked, and we were not assembling rocks from 1 particular namespace, but rather all rocks. This should be fixed now.

.github/workflows/Build-Rock.yaml
```diff
@@ -1,206 +1,227 @@
-name: Build rock
+name: Build Rock
```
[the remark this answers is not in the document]
clay-lake: Are you sure you want to change it? Usually acronyms are spelled out in all uppercase (AFAIK for example), but since we are just using it as a name, title case should be fine.

src/build_rock/configure/generate_build_matrix.py
clay-lake: If we are going to change it soon, maybe we should leave this behavior as it works in the current context.

src/build_rock/configure/generate_build_matrix.py
linostar: Didn't see any yet.

src/build_rock/configure/generate_build_matrix.py
linostar: Indeed.

.github/workflows/Build-Rock.yaml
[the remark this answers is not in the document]
linostar: I don't mind either way, but changing it here will mean you need to change it everywhere across the repo and the documentation, to keep the consistency. You better consult with Cris about that because he is the one who changed "ROCK" to "rock" a couple of months ago.

zhijie-yang (approved):
LGTM. Thanks!

.github/workflows/Build-Rock.yaml
```diff
@@ -1,54 +1,62 @@
...
+      # build configuration
       oci-archive-name:
-        description: "Final filename of the rock's OCI archive"
+        description: "Final filename of the Rock OCI archive."
         type: string
         required: true
-      oci-factory-path:
-        description: "Path, in the OCI Factory, to this rock"
+      arch-map:
+        description: "JSON string mapping target architecture to runner."
```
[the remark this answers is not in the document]
clay-lake: That is true. Outside of a x64 build, the user will have to pass a JSON string to map a runner to each additional arch. I think there is some value in being able to select particular runners for performance or load management, but lets see what I can do to automate this for most cases.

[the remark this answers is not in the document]
clay-lake: Looking at automation, this parameter is only really relevant for the LPCI fallback. We can just assume that all arches supported by runners  (x64, ARM32, ARM64) are available, then let the user reconfigure for launchpad if required. The consequence is that if the runner is not available the workflow will hang.

There are some smart ways to know ahead of time if a particular runner arch is available, but it either requires a PAT with admin ro permissions or probing the environment with some test jobs.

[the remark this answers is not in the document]
cjdcordeiro: lpci runners are not in GH, so that shouldn't matter. the information in the rockcraft.yaml should be enough to infer this info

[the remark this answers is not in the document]
cjdcordeiro: `X64` and `ARM64` runners are always available. For other rockcraft archs, yes, you can either drop support for lpci, or simply default to X64 and mark them to be built with lpci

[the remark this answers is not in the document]
clay-lake: `amd64` and `arm64` rock arches now map to linux `X64` and `ARM64` runners with labels. Unless an external user has specific self-hosted runners they want to use, they wont need to modify this parameter beyond it's current value.

For testing I removed the `arch-map` parameter from the workflow call in Image.yaml. The default value should be used now.  https://github.com/canonical/oci-factory/actions/runs/10993623923

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
...
+
+      # Job Setup
+      - uses: actions/checkout@v4
+        with:
+          repository: ${{ env.OCI_FACTORY_REPO }}
+          ref: ${{ env.OCI_FACTORY_BRANCH }}
+          fetch-depth: 1
+
+      - run: src/build_rock/assemble_rock/requirements.sh
+
```
cjdcordeiro, line 217:
```suggestion
```

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
...
+      - name: Download Single Arch Rocks
+        uses: actions/download-artifact@v4
         id: download
-      - run: sudo apt update && sudo apt install buildah -y
-      - name: Merge single-arch rocks into multi-arch OCI archive
+        with:
+          path: ${{ env.ROCK_CI_FOLDER }} 
+          pattern: ${{ inputs.oci-archive-name }}-*
+
```
cjdcordeiro, line 225:
```suggestion
```

nit: not sure if there are others like this, but better run a yaml formatter on all these files

.github/workflows/_Test-OCI-Factory.yaml
```diff
@@ -56,6 +56,9 @@ jobs:
       # test-oci-factory/pytest requirements
       - run: pip install -r tests/etc/requirements.txt
 
+      # build_rock/configure requirements 
+      - run: pip install -r src/build_rock/configure/requirements.txt
```
cjdcordeiro, line 60: Why not have this in `tests/etc/requirements.txt` too? If you want to avoid duplication it should be possible to just add `-r ../../src/build_rock/configure/requirements.txt` to that file

src/shared/github_output.py
```diff
@@ -0,0 +1,52 @@
...
+                "file_handler not available. Please use in context block."
+            )
+
+        for key, value in kwargs.items():
+
+            formatted_value = self.format_value(value)
+            print(f"{key}={formatted_value}", file=self.file_handler)
+
+    @staticmethod
+    def format_value(value):
```
cjdcordeiro, line 43: smart!

tests/data/rockcraft.yaml
```diff
@@ -0,0 +1,18 @@
...
+base: bare
+build-base: ubuntu@22.04
+platforms:
+  amd64: 
+
+parts:
+  hello:
+    plugin: nil
+    stage-packages:
+      - hello
```
cjdcordeiro, line 18:
```suggestion
      - hello

```

tests/fixtures/buffers.py
```diff
@@ -1,9 +1,25 @@
...
 
 @pytest.fixture
 def str_buff():
     """String IO fixture for simulating a file object"""
     with StringIO() as buffer:
         yield buffer
+
+
+@pytest.fixture
+def github_output(monkeypatch):
```
cjdcordeiro, line 16: you could use the `tmp_path` pytest fixture

cjdcordeiro (changes requested):
Thanks for this work @clay-lake 

It's a long review, but I don't think there are any major to controversial change requests.

One addition to this PR would be documentation: could you please add a new section to the README, explaining how one can use these workflows from within their CI? I know there's more to come (like the Test workflow), but it's easier to get the docs started as we go.

.github/workflows/Build-Rock.yaml
```diff
@@ -1,54 +1,65 @@
...
-        description: "Final filename of the rock's OCI archive"
-        type: string
-        required: true
-      oci-factory-path:
-        description: "Path, in the OCI Factory, to this rock"
-        type: string
-        required: true
-      rock-name:
-        description: "Name of the rock"
+        description: "Final filename of the Rock OCI archive."
```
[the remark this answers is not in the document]
clay-lake: Fixed that throughout the changes. Thanks :)

tests/fixtures/buffers.py
clay-lake: Fixed! Thanks :)

tests/data/rockcraft.yaml
clay-lake: Fixed, thanks :)

.github/workflows/_Test-OCI-Factory.yaml
clay-lake: Sure! That is pretty cool.

.github/workflows/Build-Rock.yaml
clay-lake: No problem. Do we have a preferred autoformatter? I can add it to the docs.

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
...
           verbosity: debug
-      - uses: actions/setup-python@v5
-        if: needs.prepare-multi-arch-matrix.outputs.build-with-lpci == 1
+
+
+      - name: Collect Artifacts
+        id: collect-artifacts
+        run: |
+              mkdir -p ${{ env.ROCK_CI_FOLDER }} && cp ${{ steps.rockcraft.outputs.rock }} "$_"
+              echo "filename=$(basename ${{ steps.rockcraft.outputs.rock }})" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
clay-lake: Should be fixed now with the auto-format.

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
...
+    # outsourced to Launchpad for completion. 
+    # Note the Secret
+    needs: [configure-build]
+    if: fromJSON(needs.configure-build.outputs.lpci-build-matrix).include[0] != ''
+    strategy: 
+      fail-fast: true
+      matrix: ${{ fromJSON(needs.configure-build.outputs.lpci-build-matrix) }}
+    runs-on: ubuntu-22.04
+    name: 'lpci-build | ${{ matrix.architecture }} '
+    steps:
```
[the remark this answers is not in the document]
clay-lake: Sounds good. It would be like some kind of adapter action: https://warthogs.atlassian.net/browse/ROCKS-1582

src/build_rock/configure/generate_build_matrix.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def configure_matrices(target_archs: list, arch_map: dict, lp_fallback: bool) -> dict:
+    """Sort build into appropriate build matrices"""
+
+    # map configuration to individual job matrices
+    build_matrices = {name.value: {"include": []} for name in MATRIX_NAMES}
+
+    # Check if we have runners for all supported architectures
+    if missing_archs := set(target_archs) - set(arch_map.keys()):
```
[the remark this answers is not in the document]
clay-lake: Thanks :)

.github/workflows/Build-Rock.yaml
```diff
@@ -1,54 +1,65 @@
...
+
+
 env:
-  ROCKS_CI_FOLDER: ci-rocks
+  ROCK_REPO_DIR: rock-repo  # path where the image repo is cloned to
+  ROCK_CI_FOLDER: ci-rocks  # path of uploaded/downloaded artifacts
+  
+  OCI_FACTORY_REPO: canonical/oci-factory
+  # TODO: update before final merge
+  OCI_FACTORY_BRANCH: ROCKS-1452/oci-factory-refactor-build-rock-to-be-externally-reusable  # required for external workflow calls.
```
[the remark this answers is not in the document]
clay-lake: My initial plan was to have a short follow up PR, where I update OCI_FACTORY_BRANCH to reflect main. But thinking back now, I think think this will make development awkward since we always reference that branch, even for testing. I think we can set OCI_FACTORY_BRANCH in such that we only enforce the default branch if the call is external, which is where we need this mechanism anyhow. 

Passing a ref in a re usable workflow seem to be a [common issue](https://github.com/actions/toolkit/issues/1264). If it is ok with you, I'll add a ticket for this, and address it in a future PR. https://warthogs.atlassian.net/browse/ROCKS-1583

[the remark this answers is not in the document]
clay-lake: I found a useful action which get around this issue, and returns the ref that was used to call the workflow. The only caveat is that it requires a GH token for private repositories.  I think we can add support for that as needed.

https://github.com/canonical/get-workflow-version-action

.github/workflows/Build-Rock.yaml
```diff
@@ -1,54 +1,65 @@
...
-      - name: Clone generic image repository
-        if: ${{ steps.clone-image-repo.outcome == 'failure' }}
-        run: |
-          git clone ${{ inputs.rock-repo }} .
+          repository: ${{ env.OCI_FACTORY_REPO }}
+          ref: ${{ env.OCI_FACTORY_BRANCH }}
+          fetch-depth: 1
 
       - name: Validate access to triggered image
         uses: ./.github/actions/validate-actor
```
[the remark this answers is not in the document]
clay-lake: Fixed. I moved the validation out to the Image workflow. The image workflow is the sole caller of  Build-Rock.yaml so it should be safe.

.github/workflows/Build-Rock.yaml
```diff
@@ -1,54 +1,65 @@
...
-        uses: actions/checkout@v4
-        id: clone-image-repo
-        continue-on-error: true
+
+      # Job Setup
+      - uses: actions/checkout@v4
         with:
-          repository: ${{ inputs.rock-repo }}
-          fetch-depth: 0
-      - name: Clone generic image repository
```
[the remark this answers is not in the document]
clay-lake: Your right. To fix this I created a new action in `.github/actions/checkout` which implements the subset of features we use in actions/checkout@v4 for an almost drop in replacement. In the future we could also add caching to the action to speed up duplicate clones within the same run.

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
           image-path: ${{ inputs.oci-factory-path }}
           github-token: ${{ secrets.ROCKSBOT_TOKEN }}
 
-      - run: git checkout ${{ inputs.rock-repo-commit }}
```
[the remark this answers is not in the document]
clay-lake: Fixed in comment above :)

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
...
+    needs: [configure-build]
+    if: fromJSON(needs.configure-build.outputs.runner-build-matrix).include[0] != ''
     strategy:
       fail-fast: true
-      matrix: ${{ fromJSON(needs.prepare-multi-arch-matrix.outputs.build-for) }}
-    runs-on: ${{ matrix.runner }}
-    name: 'Build ${{ inputs.rock-name }} | ${{ matrix.architecture }}'
+      matrix: ${{ fromJSON(needs.configure-build.outputs.runner-build-matrix) }}
+    runs-on: ${{ matrix.runner }} 
+    name: 'runner-build | ${{ matrix.architecture }} '
```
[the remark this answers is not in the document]
clay-lake: Good point. I added them back as an optional parameter to the workflow.

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
...
-      - name: Clone GitHub image repository
+
+      - name: Clone Repo
         uses: actions/checkout@v4
-        id: clone-image-repo
-        continue-on-error: true
         with:
           repository: ${{ inputs.rock-repo }}
-          fetch-depth: 0
-      - name: Clone generic image repository
```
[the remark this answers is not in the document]
clay-lake: Fixed in earlier comment on checkout action (.github/actions/checkout).

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
...
-        if: ${{ steps.clone-image-repo.outcome == 'failure' }}
-        run: |
-          git clone ${{ inputs.rock-repo }} .
-      - run: git checkout ${{ inputs.rock-repo-commit }}
```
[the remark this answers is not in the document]
clay-lake: Fixed in earlier comment on checkout action (.github/actions/checkout).

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
...
+      
+
+      - name: Clone Repo
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ inputs.rock-repo }}
+          path: ${{ env.ROCK_REPO_DIR }}
+          ref: ${{ inputs.rock-repo-commit }}
+          submodules: 'recursive'
+
```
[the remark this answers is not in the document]
clay-lake: Fixed in earlier comment on checkout action (.github/actions/checkout).

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
...
+    # configured for the required architecture. Builds in this job will be
+    # outsourced to Launchpad for completion. 
+    # Note the Secret
+    needs: [configure-build]
+    if: fromJSON(needs.configure-build.outputs.lpci-build-matrix).include[0] != ''
+    strategy: 
+      fail-fast: true
+      matrix: ${{ fromJSON(needs.configure-build.outputs.lpci-build-matrix) }}
+    runs-on: ubuntu-22.04
+    name: 'lpci-build | ${{ matrix.architecture }} '
```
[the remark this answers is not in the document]
clay-lake: I added it back as an optional workflow parameter.

.github/workflows/Build-Rock.yaml
```diff
@@ -1,54 +1,65 @@
...
+      # build parameters
       oci-archive-name:
-        description: "Final filename of the rock's OCI archive"
-        type: string
-        required: true
-      oci-factory-path:
-        description: "Path, in the OCI Factory, to this rock"
-        type: string
-        required: true
-      rock-name:
```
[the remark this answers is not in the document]
clay-lake: I moved them to the Image workflow because it seemed to have better separation of responsibilities, especially if the build-rock workflow is used for external tasks. We can keep the Build-Rock workflow for just building the rock, and the Image workflow for the factory pipeline itself.

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
           image-path: ${{ inputs.oci-factory-path }}
           github-token: ${{ secrets.ROCKSBOT_TOKEN }}
 
-      - run: git checkout ${{ inputs.rock-repo-commit }}
-
-      - run: sudo snap install yq --channel=v4/stable
-      - name: Validate image naming and base
```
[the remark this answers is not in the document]
clay-lake: Resolving for discussion above.

.github/workflows/Build-Rock.yaml
```diff
@@ -58,159 +69,171 @@ jobs:
...
+    # runner-build builds rocks per target architecture using pre configured runner images.
+    needs: [configure-build]
+    if: fromJSON(needs.configure-build.outputs.runner-build-matrix).include[0] != ''
     strategy:
       fail-fast: true
-      matrix: ${{ fromJSON(needs.prepare-multi-arch-matrix.outputs.build-for) }}
-    runs-on: ${{ matrix.runner }}
-    name: 'Build ${{ inputs.rock-name }} | ${{ matrix.architecture }}'
+      matrix: ${{ fromJSON(needs.configure-build.outputs.runner-build-matrix) }}
+    runs-on: ${{ matrix.runner }}
```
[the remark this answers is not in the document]
clay-lake: Ahh, the runner for LPCI builds is set statically in the [workflow itself](https://github.com/canonical/oci-factory/blob/43931e8400332f547098ddb7acd293c7e6d9c2fa/.github/workflows/Build-Rock.yaml#L152)

cjdcordeiro:
@clay-lake https://github.com/canonical/oci-factory/actions/runs/11342675482/job/31545731337 1.0.1 seems to be missing

clay-lake:
@cjdcordeiro I found why 1.0.1 is not building. Testing the `prepare_single_image_build_matrix.py` script (from main) seems to drop a build entirely if any of the releases in the image.yaml are EOL. Is that intended? I'll fix in another PR if not. 

Instead of testing mock-rock/1.0 here, I made a multi-arch test build in rocks-toolbox using an external workflow call.
https://github.com/canonical/rocks-toolbox/actions/runs/11365735363

cjdcordeiro:
good catch @clay-lake 

That is NOT the desired behaviour indeed. We should only ignore the track, not the whole build.

cjdcordeiro (approved):
great. thanks for the changes :) looking good. 
we need to make sure we multi-arch tests are running again so pls let me know once the fix PR for the bad EOL filtering comes up

PR #247, opened 2024-09-20
Fix CLI Client Handling Non-Interactive Terminal and Error Message for Non-Canonical Repo

zhijie-yang:

---

## Description
This PR fixes the following issues:
1. The CLI client is giving false positive results detecting non-interactive terminals.
2. The CLI client dumps stack traceback when the regex fails to match a repo remote URL that does not belong to Canonical.

### DIFF
--- a/tools/cli-client/internals/cli/cli_main.go
+++ b/tools/cli-client/internals/cli/cli_main.go
@@ -55,7 +55,7 @@ func addHelp(p *flags.Parser) {
 func blockForConfirm(s string) error {
 	// check if is a tty
 	fi, err := os.Stdin.Stat()
-	if err != nil && fi.Mode()&os.ModeNamedPipe == 1 {
+	if err != nil && fi.Mode()&os.ModeNamedPipe != 0 {
 		return fmt.Errorf("non-interactive terminal detected, run with -y option")
 	} else if err != nil {
 		return err
--- a/tools/cli-client/internals/cli/cli_main.go
+++ b/tools/cli-client/internals/cli/cli_main.go
@@ -55,7 +55,7 @@ func addHelp(p *flags.Parser) {
 func blockForConfirm(s string) error {
 	// check if is a tty
 	fi, err := os.Stdin.Stat()
-	if err != nil || fi.Mode()&os.ModeNamedPipe == 0 {
+	if err != nil && fi.Mode()&os.ModeNamedPipe == 1 {
 		return fmt.Errorf("non-interactive terminal detected, run with -y option")
 	} else if err != nil {
 		return err
--- a/tools/cli-client/internals/trigger/build_metadata.go
+++ b/tools/cli-client/internals/trigger/build_metadata.go
@@ -29,18 +29,18 @@ func InferBuildMetadata() BuildMetadata {
 	repo, err := git.PlainOpenWithOptions(".",
 		&git.PlainOpenOptions{DetectDotGit: true, EnableDotGitCommonDir: false})
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Unable to open repository: %v", err)
+		fmt.Fprintf(os.Stderr, "Unable to open repository: %v\n", err)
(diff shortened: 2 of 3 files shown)


tools/cli-client/internals/cli/cli_main.go
```diff
@@ -55,7 +55,7 @@ func addHelp(p *flags.Parser) {
 func blockForConfirm(s string) error {
 	// check if is a tty
 	fi, err := os.Stdin.Stat()
-	if err != nil || fi.Mode()&os.ModeNamedPipe == 0 {
+	if err != nil && fi.Mode()&os.ModeNamedPipe == 1 {
```
linostar, line 58:
```suggestion
	if err != nil && fi.Mode()&os.ModeNamedPipe != 0 {
```
because os.ModeNamedPipe value is higher than 1, so "bit-wise and" with it will be higher than 1 as well.

linostar (changes requested):

tools/cli-client/internals/cli/cli_main.go
zhijie-yang: Good catch. Thanks!

linostar (approved):
LGTM.

tools/cli-client/internals/trigger/build_metadata.go
```diff
@@ -49,12 +49,16 @@ func InferBuildMetadata() BuildMetadata {
 	// use regex to match the repo location
 	regex := regexp.MustCompile("github.com[:/](canonical/.*).git")
```
zhijie-yang, line 50: We should only match "Push URL: ..." (even without ".git"), because there can be the case the the developer pulls from upstream and updates the repo in canonical. (Suggested by @clay-lake )

tools/cli-client/internals/trigger/build_metadata.go
```diff
@@ -49,12 +49,16 @@ func InferBuildMetadata() BuildMetadata {
...
 	matches := regex.FindStringSubmatch(remoteURL)
+	if len(matches) < 2 {
+		fmt.Fprintf(os.Stderr, "oci-factory must be called in a git local repository under organization [canonical]\n")
+		os.Exit(1)
+	}
 	source := matches[1]
```
zhijie-yang, line 56: Then we should use `matches[0]` and `if len(matches) == 1` to match with this change.

zhijie-yang: The length of a successfully matched group should be 2, since the first match group is the whole regex, and the second match group is the `(canonical/.*)`.

tools/cli-client/internals/trigger/build_metadata.go
zhijie-yang: Updated.

tools/cli-client/internals/trigger/build_metadata.go
```diff
@@ -47,10 +47,10 @@ func InferBuildMetadata() BuildMetadata {
 	logger.Debugf("Remote URL: %s", remoteURL)
 
 	// use regex to match the repo location
-	regex := regexp.MustCompile("github.com[:/](canonical/.*).git")
+	regex := regexp.MustCompile(`github.com[:\/](canonical\/[A-Za-z0-9_-]*)(\.git)?`)
 	matches := regex.FindStringSubmatch(remoteURL)
-	if len(matches) < 2 {
-		fmt.Fprintf(os.Stderr, "oci-factory must be called in a git local repository under organization [canonical]\n")
+	if len(matches) < 3 {
+		fmt.Fprintf(os.Stderr, "oci-factory must be called in a git local repository belongs to the organization [canonical]\n")
```
cjdcordeiro, line 53:
```suggestion
		fmt.Fprintf(os.Stderr, "oci-factory must be called in a git local repository belonging to the organization [canonical]\n")
```

cjdcordeiro (approved):

PR #290, opened 2024-11-20
Add CI-Level trivyignore

zhijie-yang:

---

## Description
This PR adds a CI-level trivyignore to bypass the false positive vulnerability findings for the pebble binary in the bare-based rocks using the feature that `trivyignores` can be a comma-separated string to contain multiple bypass lists.

### Related issues
https://github.com/canonical/chiselled-python/issues/26

### DIFF
--- a/.github/.trivyignore
+++ b/.github/.trivyignore
@@ -1,0 +1,3 @@
+# False positive for encoding/gob that not used by pebble
+# encoding/gob: cover missed cases when checking ignore depth
+CVE-2024-34156
--- a/.github/workflows/Vulnerability-Scan.yaml
+++ b/.github/workflows/Vulnerability-Scan.yaml
@@ -87,13 +87,10 @@ jobs:
       - name: Check for .trivyignore
         id: trivyignore
         run: |
+          file="./.github/.trivyignore"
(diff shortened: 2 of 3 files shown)


cjdcordeiro (approved):
LGTM. Question for future selves: how would we know when to drop this global ignore?

zhijie-yang:
If trivy still gives this "false positive" based on the Go version that pebble is built upon, then we have to wait until pebble is built upon the updated Go versions where this CVE is fixed, or the behaviour of trivy is optimised for vulnerability identification with Go binaries.

zhijie-yang:
We can add a job in the continuous testing workflow, to test if we can pass the pipeline without the bypass list, and if this is the case, the workflow can create an issue in OCI-Factory to notice that we can remove the bypass list item.

cjdcordeiro:
it's an idea to be explored, yes. It should also cover the remaining rocks' trivyignore files. let's add it to the backlog

zhijie-yang:
We have another approach in #291. I think #291 is a more decent solution for the pebble issue for now.

cjdcordeiro:
Fully agree! It was indeed already there, we just replaced `pebble` :p 

feel free to close this

PR #295, opened 2024-11-22
Issue for distinct revisions when vulnerabilities are found

zhijie-yang:

---

## Description
1. GitHub issues will be created w.r.t. each revision of an image that contains CVEs found in the vulnerability scanning workflows triggered with continuous testing workflow. In the report, all the affected tags associated with the revision will be included in the report.
2. When vulnerabilities are found for the images tested in the `Image` workflow triggered by a push, a pull request or a workflow dispatch event, the vulnerability report will be dumped to a summary instead of creating an issue at the image repo.
3. This shall fix the issue that an issue will be accidentally closed when the different risks in a track of an image are tagged to different revisions, and only some of the revisions have found CVEs.

## Tests
- Workflow summary: https://github.com/canonical/oci-factory/actions/runs/11973639728#summary-33383300917
- Issue for distinct revision: https://github.com/canonical/traefik-rock/issues/61


src/shared/release_info.py
```diff
@@ -82,3 +85,66 @@ def get_revision_to_track(all_revisions_tags: list) -> dict:
...
+
+    for revision, tags in revision_to_released_tags.items():
+        revision_to_released_tags[revision] = sorted(tags)
+
+    return dict(revision_to_released_tags)
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
```
clay-lake, line 128: I think we should create an individual store_true flag to execute get_revision_to_released_tags. We could even make it mutually exclusive of other "command" flags if required. That way it is integrated into the --help info.

tests/unit/test_shared_release_info.py
```diff
@@ -0,0 +1,67 @@
...
+            "candidate": {"target": "1.19.0-22.04_beta"},
+            "beta": {"target": "1.19.0-22.04_candidate"},
+            "edge": {"target": "5"},
+        }
+    }
+
+    with pytest.raises(BadChannel) as excinfo:
+        get_revision_to_released_tags(all_releases)
+    assert (
+        "Tag 1.19.0-22.04_candidate was caught in a circular dependency, following tags that follow themselves. Cannot pin a revision."
```
clay-lake, line 40: nit:   I think we over constraining the tests to check for this string. Maybe it is good enough that we assert BadChannel is raised. We could also then create a new exception class for circular dependencies if we need to single out this exact scenario.

clay-lake (changes requested):

.github/workflows/Vulnerability-Scan.yaml
```diff
@@ -203,11 +208,19 @@ jobs:
     if: ${{ !cancelled() && github.event_name != 'pull_request' }}
     steps:
       - uses: actions/checkout@v4
+      
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
+
+      - run: pip install pydantic==2.8.2
```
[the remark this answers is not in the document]
zhijie-yang: We can, but they contain many things beyond the requirements for this job.

.github/workflows/Vulnerability-Scan.yaml
```diff
@@ -27,8 +27,8 @@ on:
         required: false
         type: string
         default: '9999-12-31T23:59:59'
-      is-from-release:
-        description: 'If the image is from a release, we create issues when vulnerabilities are found'
+      create-issue:
+        description: 'If to create a GitHub issues for found vulnerabilities'
```
SamirPS, line 31: you can remove the If :)

clay-lake (approved):
There a few things we should work on later, but they might be out of scope of this PR. 

This is a nice feature! :+1:

PR #308, opened 2024-12-10
fix: release identification in prepare_single_image_build_matrix.py

clay-lake:
---

Included in this PR:
- fix for determining if there are builds to execute for image trigger files
- tests to check behavior of this change

### DIFF
--- a/src/image/prepare_single_image_build_matrix.py
+++ b/src/image/prepare_single_image_build_matrix.py
@@ -107,11 +107,20 @@ def filter_eol_tracks(build: dict[str, Any]) -> dict[str, Any]:
 
 
 def filter_eol_builds(builds: list[dict[str, Any]]) -> list[dict[str, Any]]:
-    """Remove any builds with no tracks specified."""
-    # remove any end of life tracks
-    builds = [filter_eol_tracks(build) for build in builds]
+    """Remove any builds with eol tracks."""
 
-    return [build for build in builds if len(build["release"])]
+    # if no release exists and therefore no eol is specified, do nothing
+    non_release_builds = [build for build in builds if "release" not in build]
+
+    # if we have release info, then filter based on eol
+    release_builds = [
+        filtered_build
+        for build in builds
+        if "release" in build
+        and (filtered_build := filter_eol_tracks(build))["release"]
+    ]
+
+    return non_release_builds + release_builds
 
 
 def write_revision_data(data_dir: Path, build: dict[str, Any]):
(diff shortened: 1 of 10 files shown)


src/image/prepare_single_image_build_matrix.py
```diff
@@ -228,7 +228,7 @@ def main():
         # so let's remove this field since we don't need it for the builds
         del build["release"]
 
-    release_to = "true" if "release" in image_trigger else ""
+    release_to = "true" if builds else ""
```
cjdcordeiro, line 231: I don't think this is the desired behavior. We don't release every time we build. You may want to build a rock without releasing it (thus staying just in GHCR)

cjdcordeiro (changes requested):

tests/data/image_without_release.yaml
```diff
@@ -0,0 +1,37 @@
...
+      1-22.04:
+        end-of-life: "2030-05-01T00:00:00Z"
+        risks:
+          - candidate
+          - edge
+          - beta
+  - source: "canonical/rocks-toolbox"
+    commit: 17916dd5de270e61a6a3fd3f4661a6413a50fd6f
+    directory: mock_rock/1.2
+    release:
```
cjdcordeiro, line 33: as per my comment above, you may want to have one of these tests have no `release` fields at all, since those are optional

src/image/prepare_single_image_build_matrix.py
cjdcordeiro: why not

```python
    release_to = "true" if "release" in image_trigger else ""

    for build in builds:
        write_revision_data(args.revision_data_dir, build)
        # the workflow GH matrix has a problem parsing nested JSON dicts
        # so let's remove this field since we don't need it for the builds
        if "release" in build:
            release_to = "true"
            del build["release"]
```

clay-lake: Right! I see what you mean now.

clay-lake: I just pushed some changes fix that. We also use a bool for release_to for clarity.

tests/data/image_without_release.yaml
clay-lake: Makes sense. I added additional tests for the image_trigger file which should cover most of the cases.

zhijie-yang (changes requested):
Thanks for the changes. They look nice to me. However, we should also cover the tests against the `release` field in the revision data generated by `prepare_single_image_build_matrix.py`, to make sure only the images we want to release has the non-empty `release` field in their revision data.

clay-lake:
Added additional testing to check how many releases there should be per image_trigger.yaml sample.

PR #342, opened 2025-02-04
Onboard istio ztunnel

IbraAoad:

---

## Description
This PR adds the ztunnel ROCK to the OCI factory.

### Related issues
https://github.com/canonical/oci-factory/issues/317


zhijie-yang (changes requested):
Thanks for the code.

I have to manually terminate the workflow runs for the black-box testing, as this rock is causing infinite loops there.

The is due to using `ztunnel` as the `entrypoint-service` rather than the default `pebble`, causing the test to fail to run the command `pebble help`, while having `pebble` indefinitely trying to run the failing `ztunnel [ ]` command.

Besides, using the programs other than `pebble` as the entry point service does not comply with the guidelines to build rocks. You can observe the warnings when running `rockcraft pack`:

```
Warning: defining an entrypoint-service will result in a rock with an atypical OCI Entrypoint. While that might be acceptable for testing and personal use, it shall require prior approval before submitting to a Canonical registry namespace.
```

I believe it is necessary to change the entry point service of this rock. Otherwise, please provide justifications for doing so. 

Discussions are welcomed. Cheers!

simskij:
> Thanks for the code.
> 
> I have to manually terminate the workflow runs for the black-box testing, as this rock is causing infinite loops there.
> 
> The is due to using `ztunnel` as the `entrypoint-service` rather than the default `pebble`, causing the test to fail to run the command `pebble help`, while having `pebble` indefinitely trying to run the failing `ztunnel [ ]` command.
> 
> Besides, using the programs other than `pebble` as the entry point service does not comply with the guidelines to build rocks. You can observe the warnings when running `rockcraft pack`:
> 
> ```
> Warning: defining an entrypoint-service will result in a rock with an atypical OCI Entrypoint. While that might be acceptable for testing and personal use, it shall require prior approval before submitting to a Canonical registry namespace.
> ```
> 
> I believe it is necessary to change the entry point service of this rock. Otherwise, please provide justifications for doing so.
> 
> Discussions are welcomed. Cheers!

We cannot get rid of the entrypoint service, as Istio launches the container with arguments that are resolved at deploy-time. As the rock needs to be interchangeable with the upstream OCI, we can't really change this without making the rock unusable outside of the charm context.

zhijie-yang:
@cjdcordeiro Need your input for the same question as https://github.com/canonical/oci-factory/pull/345#issuecomment-2642446724

cjdcordeiro (changes requested):
ACK. If it cannot be avoided, then let's go with it, but let's pls update the documentation to reflect this rock's UX.

IbraAoad:
@zhijie-yang @cjdcordeiro We managed to drop the `entrypoint service` from the rock in the latest bump and PR should be updated now, could you guys have another look to confirm that this is okay to merge?

oci/istio-ztunnel/documentation.yaml
```diff
@@ -0,0 +1,20 @@
+version: 1
+# --- OVERVIEW INFORMATION ---
+application: ztunnel
+description: >
+  Ztunnel is a purpose built implementation of the node proxy 
+  to be used as the ztunnel component of Istio's ambient mesh.
+  Read more on the [project repo](https://github.com/istio/ztunnel).
+
```
zhijie-yang, line 8: Please add the section for `parameters` to indicate to the users how to deploy this container image. If this container image is not intended to be manually deployed, please leave a note in the description.

zhijie-yang (changes requested):
Code looks nice, and all CI checks are passing, great! Here is a nitpick on the documentation:

oci/istio-ztunnel/documentation.yaml
IbraAoad: Done, the only required param is a sub-command `proxy` which is baked by default into the rock, but I added it into the `documentation.yaml` for reference

zhijie-yang: Many thanks!

zhijie-yang (approved):
LGTM, thanks.

PR #349, opened 2025-02-07
stable release for valkey with LTS support

reneradoi:
Update the channel tag for Valkey from `edge` to `stable` and update the support period according to the LTS of 24.04.

### DIFF
--- a/oci/valkey/image.yaml
+++ b/oci/valkey/image.yaml
@@ -6,6 +6,6 @@ upload:
     directory: .
     release:
       7.2.7-24.04:
-        end-of-life: "2025-12-01T00:00:00Z"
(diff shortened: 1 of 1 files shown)


zhijie-yang (approved):
LGTM, thanks.

cjdcordeiro:
@zhijie-yang @reneradoi **urgent**: this was merged with an EOL pointing to 2034 - 10 years

that falls into ESM, and thus not published in this Docker Hub's and ECR's namespace

reneradoi:
This had been previously confirmed before creating this PR. I see there is already a change being proposed by @Mehdi-Bendriss , so I guess it should be addressed.

PR #365, opened 2025-02-21
adding pilot rock

IbraAoad:

---

## Description
This PR adds the istio-pilot ROCK to the OCI factory.

## Important Note
This Rock uses an `entrypoint service`, as Istio launches the container with arguments that are resolved at deploy-time. This the rock needs to be interchangeable with the upstream OCI, we can't really change this without making the rock unusable outside of the charm context.

To tackle I have included all subcommands/flags that could be passed when using this rock along with their default values in the rock's `documentation.yaml`.

### Related issues
#361

### DIFF
--- a/oci/istio-pilot/contacts.yaml
+++ b/oci/istio-pilot/contacts.yaml
@@ -1,0 +1,5 @@
+notify:
+  emails:
+    - observability@lists.launchpad.net
+  mattermost-channels:
+    - 1ayd5kim67bbing34i3h1x9uac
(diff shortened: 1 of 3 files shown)


cjdcordeiro:
at first glance, this lgtm. I'd like to point @rbenour here cause this is an interesting documentation point to be taken for the ongoing revamping of the docs -> _rocks with an `entrypoint-service`_ (https://documentation.ubuntu.com/rockcraft/en/stable/reference/rockcraft.yaml.html#entrypoint-service)

LP seems unavailable atm so the Documentation workflow would fail. let's wait for it to be back up.

PR #387, opened 2025-03-07
feat: add chiselled JRE 21 to oci-factory

vpa1977:

---

## Description

This PR adds chiselled image for OpenJDK 21 JRE to the OCI factory.

### Related issues

https://github.com/canonical/oci-factory/issues/386

---

*Picture of a cool rock:*


zhijie-yang:
The failed efficiency test seems to result from the behaviour of `dive` and `rockcraft`.

When `run-user` is specified in the `rockcraft.yaml`, `rockcraft` will add a new layer to the image to modify the `/etc/group` and `/etc/passwd`. This works fundamentally correct.

When `dive` calculates the `userWastedPercent`, the size of the base image (the first layer) is disregarded. This is also fundamentally correct for non-baseless images.

However, in this case, we are inspecting a bare-based image. This image has three layers: The first layer contains all the contents in the `parts` from the `rockcraft.yaml`; The second layer contains the changes regarding `run-user`; The third layer contains the control files of the rock. The sizes of the three layers are by the magnificence of tens or hundreds of MBs, KBs, and KBs respectively. Since the file sizes of `/etc/{group,passwd}` in the second layer are not negligible when `base-passwd_data` is included as one of the `stage-packages` in the `parts`, the calculation of `userWastedPercent` using the last two layers can easily explode the threshold.

It turns out that we need some more sophisticated workarounds for this issue when testing the image efficiency. Looping @cjdcordeiro in for further discussions.

vpa1977:
> The failed efficiency test seems to result from the behaviour of `dive` and `rockcraft`.
> 
> When `run-user` is specified in the `rockcraft.yaml`, `rockcraft` will add a new layer to the image to modify the `/etc/group` and `/etc/passwd`. This works fundamentally correct.
> 
> When `dive` calculates the `userWastedPercent`, the size of the base image (the first layer) is disregarded. This is also fundamentally correct for non-baseless images.
> 
> However, in this case, we are inspecting a bare-based image. This image has three layers: The first layer contains all the contents in the `parts` from the `rockcraft.yaml`; The second layer contains the changes regarding `run-user`; The third layer contains the control files of the rock. The sizes of the three layers are by the magnificence of tens or hundreds of MBs, KBs, and KBs respectively. Since the file sizes of `/etc/{group,passwd}` in the second layer are not negligible when `base-passwd_data` is included as one of the `stage-packages` in the `parts`, the calculation of `userWastedPercent` using the last two layers can easily explode the threshold.
> 
> It turns out that we need some more sophisticated workarounds for this issue when testing the image efficiency. Looping @cjdcordeiro in for further discussions.

As a workaround for the time-being - should I set up the user inside the part, rather than use the run-user?

cjdcordeiro:
> When dive calculates the userWastedPercent, the size of the base image (the first layer) is disregarded. This is also fundamentally correct for non-baseless images.

@zhijie-yang is this true?

I think the problem is simpler than that:
 - /etc/{passwd,group} are present in 2 different layers, thus the inefficiency (cause they are useless in the first layer where they appear - just taking space)

cjdcordeiro:
@vpa1977 I think you can safely remove `base-passwd_data` in this case. Otherwise, @zhijie-yang , are there any adjustments we can do to `dive` to make it more permissive to such cases?

zhijie-yang:
I don't think there are any ways we can adjust how `dive` calculates `userWastedPercent`. Let's see if @vpa1977 can verify it's safe to remove `base-passwd_data` from the parts.

vpa1977:
> I don't think there are any ways we can adjust how `dive` calculates `userWastedPercent`. Let's see if @vpa1977 can verify it's safe to remove `base-passwd_data` from the parts.

Yes, it is safe to remove, `run_user` tag adds a user. The pipeline passes now.

vpa1977:
I've sent a email to Security Team to enable monitoring for `ubuntu/jre` images

cjdcordeiro (changes requested):
Great! All good from a rock's perspective.

We need some adjustments to the docs though @vpa1977 .

Since this is going to release a new tag that is a rock, whilst older tags are Docker images, we must update the docs accordingly, highlighting the change and providing the usage guides for both.

This is the first rock undergoing such transition, so while better docs come into place, I suggest updating the Usage section, with something like:

```md
From version 21 onwards...
(examples)

For all other older versions...
(examples)
```

Also, remember that your current jre image in DH has some nice examples already, which would be lost with the documentation.yaml file in this PR

vpa1977:
> Also, remember that your current jre image in DH has some nice examples already, which would be lost with the documentation.yaml file in this PR

I have updated the documentation merging the readme from the DH.

oci/jre/documentation.yaml
```diff
@@ -0,0 +1,132 @@
...
+
+  ## Usage
+
+  Launch this image locally:
+
+  For versions 8 and 17
+  ```bash
+  docker run -d --name jre-container -e TZ=UTC ubuntu/jre:17-22.04_edge
+  ```
+  For version 21
```
cjdcordeiro, line 61:
```suggestion
  For versions equal to or greater than 21
```

vpa1977: That's not factually true - the next thing we will merge - openjdk-11 rock.

cjdcordeiro: hum I see...then maybe change the other line below (https://github.com/canonical/oci-factory/pull/387/files/d184ca341d9bd880234c16013bf49b1de2d652f5#diff-cf869e26197ee838cc34353f884ae9cf4f4cd3d0b7486700b918fbcb674adeddR128) that says "from 21 onwards)

vpa1977: > hum I see...then maybe change the other line below (https://github.com/canonical/oci-factory/pull/387/files/d184ca341d9bd880234c16013bf49b1de2d652f5#diff-cf869e26197ee838cc34353f884ae9cf4f4cd3d0b7486700b918fbcb674adeddR128) that says "from 21 onwards)

done

cjdcordeiro: thanks

PR #437, opened 2025-04-14
nginx prom exporter oci v1

PietroPasotti:

---

## Description
Adds a release of the nginx-prometheus-exporter rock to OCI factory

### Related issues
#410 

---

*Picture of a cool rock:*
image

### DIFF
--- a/oci/nginx-prometheus-exporter/documentation.yaml
+++ b/oci/nginx-prometheus-exporter/documentation.yaml
@@ -11,14 +11,14 @@ description: >
 docker:
   parameters:
     - -p 9113:9113
-  access: The metrics endpoint is available at `http://localhost:9113/metrics` 
+  access: Expose the metrics endpoint at `http://localhost:9113/metrics` 
 parameters:
   - type: -e
     value: 'TZ=UTC'
     description: Timezone.
   - type: -p
     value: '9113:9113'
-    description: Expose the metrics endpoint on `localhost:9113/metrics`.
+    description: Expose the metrics endpoint at `http://localhost:9113/metrics`.
 debug:
   text: |
     ### Debugging
--- a/oci/nginx-prometheus-exporter/documentation.yaml
+++ b/oci/nginx-prometheus-exporter/documentation.yaml
@@ -1,0 +1,33 @@
+version: 1
+# --- OVERVIEW INFORMATION ---
+application: nginx-prometheus-exporter
+description: >
+  Nginx prometheus-exporter is a Prometheus exporter for Nginx.
+  Read more in the [official documentation](https://github.com/nginx/nginx-prometheus-exporter)
+  Please note that this repository holds a rock instead of a Dockerfile-based image.
+  As such, the entrypoint is Pebble.
+  Read more in the [Rockcraft docs](https://canonical-rockcraft.readthedocs-hosted.com/en/latest/).
+# --- USAGE INFORMATION ---
+docker:
+  parameters:
+    - -p 9113:9113
+  access: The metrics endpoint is available at `http://localhost:9113/metrics` 
+parameters:
+  - type: -e
+    value: 'TZ=UTC'
+    description: Timezone.
+  - type: -p
+    value: '9113:9113'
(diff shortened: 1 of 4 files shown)


oci/nginx-prometheus-exporter/documentation.yaml
```diff
@@ -0,0 +1,33 @@
...
+  Nginx prometheus-exporter is a Prometheus exporter for Nginx.
+  Read more in the [official documentation](https://github.com/nginx/nginx-prometheus-exporter)
+  Please note that this repository holds a rock instead of a Dockerfile-based image.
+  As such, the entrypoint is Pebble.
+  Read more in the [Rockcraft docs](https://canonical-rockcraft.readthedocs-hosted.com/en/latest/).
+# --- USAGE INFORMATION ---
+docker:
+  parameters:
+    - -p 9113:9113
+  access: The metrics endpoint is available at `http://localhost:9113/metrics`
```
zhijie-yang, line 14: Can you unify this description with the one down in the parameters?

zhijie-yang (changes requested):
This looks good to me, thanks! Just one nitpick:

zhijie-yang:
It seems we have an updated CVE overnight. @PietroPasotti, please add it to the `.trivyignore` file with justification, or patch it in the build recipe?

https://github.com/canonical/oci-factory/actions/runs/14463198878/attempts/1#summary-40559799701

PietroPasotti:
we can't ignore it as it does affect the image. It's already fixed in the upstream but they didn't release it yet, we're trying to understand what to do...

zhijie-yang:
I see it. If we are not in a hurry to publish this image, we can wait a bit until they release the new version. Otherwise, you can add this CVE to the `.trivyignore` to let the CI pass, but add a reminder in your backlog to update to the latest release when it's available and remove this CVE from `.trivyignore`. 

@cjdcordeiro wdyt?

cjdcordeiro:
it's a product decision from @PietroPasotti 's team. 

If our teams aren't maintaining that source, that we can't commit to support it, and thus it will always be prone to CVEs (i.e. even if there's a fix for the current CVE, tomorrow there will be another CVE and the image would have already been published).

This is why we've been advising Maintainers to set super short EOLs for such images. Ideally, they should be onboarded to our security team for proper support.

simskij:
@PietroPasotti There are zero code paths within the upstream source code that utilize the affected module `golang.org/x/crypto/ssh`. As such, this is to be considered a false positive and added to trivy's ignore list.

zhijie-yang:
In this case, I suggest that @PietroPasotti can provide a scan result from `govulncheck` as more evident justification, like [the one for `parca`](https://github.com/canonical/parca-rock/issues/6#issuecomment-2812277725). We can then safely add them to the ignore list.

simskij:
> In this case, I suggest that @PietroPasotti can provide a scan result from `govulncheck` as more evident justification, like [the one for `parca`](https://github.com/canonical/parca-rock/issues/6#issuecomment-2812277725). We can then safely add them to the ignore list.

Here you go. The only listed one is a medium. 
I'll ask Pietro to add a trivyignore.
```
govulncheck ./   
=== Symbol Results ===

Vulnerability #1: GO-2025-3503
    HTTP Proxy bypass using IPv6 Zone IDs in golang.org/x/net
  More info: https://pkg.go.dev/vuln/GO-2025-3503
  Module: golang.org/x/net
    Found in: golang.org/x/net@v0.34.0
    Fixed in: golang.org/x/net@v0.36.0
    Example traces found:
      #1: exporter.go:250:47: nginx.registerCollector calls client.NewNginxClient, which eventually calls httpproxy.config.useProxy

Your code is affected by 1 vulnerability from 1 module.
This scan also found 0 vulnerabilities in packages you import and 3
vulnerabilities in modules you require, but your code doesn't appear to call
these vulnerabilities.
Use '-show verbose' for more details.
```

PietroPasotti:
[like so](https://pastebin.canonical.com/p/BfVrw75xWD/)?

zhijie-yang (approved):
This looks nice. Thanks!

PR #469, opened 2025-06-02
ci: decouple vulnerability-scan workflow from oci-factory

alesancor1:
The input required only for OCI-Factory are now optional. A check has been added for it.

---

## Description
Modified the CVE Scanning workflow to make it not depend on OCI-Factory. Now it can be used by other repositories by providing just the image to be scanned.

A new check has been added for the `oci-image-path` input to make sure it is required when the workflow is run in OCI Factory repository, and the `notify` job will only run when the OCI Factory is triggering the workflow.


zhijie-yang (approved):
Looks good to me. Thanks!

PR #475, opened 2025-06-06
ci(Build-Rock): add flag to run rockcraft test conditionally

alesancor1:

---

## Description

A new flag to the `rockcraft pack` action will be added to conditionally run `rockcraft test`. This PR adapts the current reusable `Build-Rock` workflow to cope with the new flag.

Blocked by: https://github.com/canonical/craft-actions/pull/40

### DIFF
--- a/.github/workflows/Build-Rock.yaml
+++ b/.github/workflows/Build-Rock.yaml
@@ -11,6 +11,9 @@ on:
       build-id:
         description: "Optional string for identifying workflow jobs in GitHub UI"
         type: string
+      rockcraft-test:
+        description: "Whether to run rockcraft test when packing the rock"
+        type: boolean
 
       # source parameters
       rock-repo:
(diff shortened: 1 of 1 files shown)


.github/workflows/Build-Rock.yaml
```diff
@@ -11,6 +11,9 @@ on:
       build-id:
         description: "Optional string for identifying workflow jobs in GitHub UI"
         type: string
+      rockcraft-test:
+        description: "Whether to run rockcraft test when packing the rock"
+        type: boolean
```
cjdcordeiro, line 16: this means it's mandatory, but it's optional in the action. wanna default to `false`?

cjdcordeiro (changes requested):

.github/workflows/Build-Rock.yaml
alesancor1: GH Actions [docs](https://docs.github.com/en/actions/reference/metadata-syntax-for-github-actions#inputsinput_idrequired) suggest that `required` is false by default, as it is only necessary to specify it when something is actually required (i.e set to `true`). So it should work anyway, as the action itself will consider `test: false` if no value is passed down to it.

In any case, I added the default value just to be safe :+1: 
Check: https://github.com/canonical/oci-factory/pull/475/commits/8d319ed030897d08ac8630a15adb0e14178e5045

PR #494, opened 2025-07-03
fix(ci): have the rocksbot force push to bypass branch protection

cjdcordeiro:
This can only be properly tested in `main`, since that's where the branch protection rules are constraining our workflows


alesancor1 (approved):

zhijie-yang (approved):
LGTM, worth a try.

PR #519, opened 2025-07-17
feat: add dotnet-runtime:9.0-25.04_edge

zhijie-yang:

---

## Description

This PR adds the dotnet-runtime:9.0-25.04_edge.

The documentation is handled similarly to the `jre` rock to cope with the co-existence of Dockerfile-based images and rocks.

### Related issues

#517

### DIFF
--- a/oci/dotnet-runtime/documentation.yaml
+++ b/oci/dotnet-runtime/documentation.yaml
@@ -145,3 +145,9 @@ debug:
     ```bash
     docker logs -f dotnet-runtime-container
     ```
+
+    For versions 9.0-25.04 and onward, to inspect application logs:
+
+    ```bash
+    docker exec -it dotnet-runtime-container pebble logs
+    ```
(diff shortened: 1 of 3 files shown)


oci/dotnet-runtime/documentation.yaml
cjdcordeiro, line 1: this is the entire md, which will then render duplicated content. pls fix the description

cjdcordeiro (changes requested):

oci/dotnet-runtime/documentation.yaml
zhijie-yang: The documentation for the [jre rock](https://github.com/canonical/oci-factory/blob/main/oci/jre/documentation.yaml) is also doing this. Wouldn't this be the only way to keep the documentation for both the dockerfile-based image and the rock?

zhijie-yang: Updated.

PR #526, opened 2025-07-21
fix: modify nodejs tag

alesancor1:

---

## Description
Modify the nodejs tag


cjdcordeiro:
P.S> the Notify CI issue is unrelated, but needs taking care of

zhijie-yang (changes requested):
With an EOL in the past, I don't think this image will be built and released, so that new tag will not get to the registries. I’d suggest you put an EOL in the very near future (say tomorrow 0:00) to let the CI run.

zhijie-yang (approved):
Looks good to me. Thanks!

PR #528, opened 2025-07-22
feat: add MySQL .trivyignore file

sinclert-canonical:

---

## Description
This PR adds a `.trivyignore` file to the MySQL OCI Factory so that the false-positive vulnerabilities preventing the automatic publishing of the image are ignored (see [CI run](https://github.com/canonical/oci-factory/actions/runs/16118774952/)).

### DIFF
--- a/oci/mysql/.trivyignore
+++ b/oci/mysql/.trivyignore
@@ -1,11 +1,17 @@
 # Golang CVE fixed in upstream PR
 # https://go-review.googlesource.com/c/go/+/578375
+# Patched version released in Ubuntu archive
+# https://ubuntu.com/security/CVE-2024-24788
 CVE-2024-24788
 
 # Golang CVE fixed in upstream PR
 # https://go-review.googlesource.com/c/go/+/590296
+# Patched version released in Ubuntu archive
+# https://ubuntu.com/security/CVE-2024-24790
 CVE-2024-24790
 
 # Golang CVE fixed in upstream PR
 # https://go-review.googlesource.com/c/go/+/611239
+# Patched version released in Ubuntu archive
+# https://ubuntu.com/security/CVE-2024-34156
 CVE-2024-34156


zhijie-yang:
I have a question regarding the customized image trigger: Why the tack to be releases is `.-` instead of `8.0-24.04`? Besides, if the CVE fixes have been patched and reflected in the rockcraft build recipe, you shouldn't need to add them to the `.trivyignore` file.

sinclert-canonical:
> I have a question regarding the customized image trigger: Why the tack to be releases is `.-` instead of `8.0-24.04`? Besides, if the CVE fixes have been patched and reflected in the rockcraft build recipe, you shouldn't need to add them to the `.trivyignore` file.

Thanks for the quick reply.

I do not fully understand the 1st question: I believe this repository supports the specification of images with the `<workload-major>.<workload-minor>-<ubuntu-major>.<ubuntu-minor>` format, similar to what other Data Platform products have (see [Kafka](https://github.com/canonical/oci-factory/blob/35c9cdfe15d3d71953f26a757f174868da528a5e/oci/kafka/image.yaml#L8), [Valkey](https://github.com/canonical/oci-factory/blob/35c9cdfe15d3d71953f26a757f174868da528a5e/oci/valkey/image.yaml#L8), etc).

As for the 2nd question: judging by the Go upstream PRs, these vulnerabilities have been fixed in old versions of Golang 1.22. Latest 1.22, 1.23 and 1.24 versions should not be impacted by such issues. AFAIK, that is why we consider them false positives. cc @paulomach

zhijie-yang:
The [custom trigger](https://github.com/canonical/oci-factory/actions/runs/16118774952/job/45478914414#step:7:1) specifies a track `.-`. Run this command you'll obtain
```
version: 1
upload:
    - source: canonical/mysql-rock
      commit: 3c55611b6a9f4c5264c289fc414aed33065fb285
      directory: ./
      release:
        .-:
            end-of-life: "2030-07-07T00:00:00Z"
            risks:
                - edge

```

sinclert-canonical:
> The [custom trigger](https://github.com/canonical/oci-factory/actions/runs/16118774952/job/45478914414#step:7:1) specifies a track `.-`.

Oh, that I have no clue. I see other teams specifying incremental versions of their workload (i.e. `X-24.04`, `X.Y-24.04`, etc), maybe they were not sure how many version parts do the pipelines support? 🤷🏻‍♂️

zhijie-yang:
The track has to be of format `<version>-<base>` per the [documentation](https://github.com/canonical/oci-factory?tab=readme-ov-file#trigger-files). I think the cause of this weird track name comes from `the workflow in the mysql-rock repo`(https://github.com/canonical/mysql-rock/actions/runs/16443490426/job/46469869051#step:6:10)

sinclert-canonical:
> The track has to be of format `<version>-<base>` per the [documentation](https://github.com/canonical/oci-factory?tab=readme-ov-file#trigger-files). I think the cause of this weird track name comes from [the workflow in the mysql-rock repo](https://github.com/canonical/mysql-rock/actions/runs/16443490426/job/46469869051#step:6:10)

Interesting... It seems that accessing elements of a `yq` split sequence via `.<number>` is not valid with the version that gets installed via Debian package registry. When installing from source, as in [this workflow](https://github.com/canonical/identity-team/blob/4c283f70c418d7b247e3cdab89934176d8ed9289/.github/workflows/_rock-oci-publish.yaml#L46-L58) from the Identity team, that is valid syntax. I will tweak our private repository to use `[<number>]` instead of `.<number>`.

zhijie-yang:
Hey @sinclert-canonical , I don't think the track `'"8"."0"-"24.04"'` is something you want to have :joy:

sinclert-canonical:
> Note: These CVE findings are already patched for the `gosu` in Ubuntu archive by rebuilding with the `golang-1.22` package that receives the security patchings (version >=1.22.2-2ubuntu0.3).

We are using `ubuntu@24.04` as the rock base (see [code](https://github.com/canonical/mysql-rock/blob/7f0a47e5b648b9183b952bf5fbb8f095b3b23780/rockcraft.yaml#L2)), which, according to the links you provided, already has the fixed Golang 1.22. If so, why is the [CI run](https://github.com/canonical/oci-factory/actions/runs/16468600687) still complaining about vulnerabilities? I think this discrepancy is the reason why my manager suspected the vulnerability reports are _false-positives_. Are we missing something?

oci/mysql/.trivyignore
```diff
@@ -0,0 +1,11 @@
+# Golang CVE fixed in upstream PR
+# https://go-review.googlesource.com/c/go/+/578375
```
zhijie-yang, line 2:
```suggestion
# https://go-review.googlesource.com/c/go/+/578375
# Patched version released in Ubuntu Archive
# https://ubuntu.com/security/CVE-2024-24788
```

oci/mysql/.trivyignore
```diff
@@ -0,0 +1,11 @@
...
+CVE-2024-24788
+
+# Golang CVE fixed in upstream PR
+# https://go-review.googlesource.com/c/go/+/590296
```
zhijie-yang, line 6:
```suggestion
# https://go-review.googlesource.com/c/go/+/590296
# Patched version released in Ubuntu Archive
# https://ubuntu.com/security/CVE-2024-24790
```

oci/mysql/.trivyignore
```diff
@@ -0,0 +1,11 @@
...
+CVE-2024-24790
+
+# Golang CVE fixed in upstream PR
+# https://go-review.googlesource.com/c/go/+/611239
```
zhijie-yang, line 10:
```suggestion
# https://go-review.googlesource.com/c/go/+/611239
# Patched version released in Ubuntu Archive
# https://ubuntu.com/security/CVE-2024-34156
```

zhijie-yang (commented):
I'd have one more nitpick on the justifications for the trivyignore file, and we are good to go then!

zhijie-yang (approved):

PR #549, opened 2025-08-05
fix(ci): checkout rock repo for trivyignore files

### LINKED ISSUE #524: bug: trivyignore-path doesn't work when the workflow is called from outside the oci-factory
https://github.com/canonical/oci-factory/blob/4562eface85d08b6d377a9419e83ea748703337b/.github/workflows/Test-Rock.yaml#L38

The job where this file is used only checks out the oci-factory repo, and not the original repo that called the workflow, so passing this input to the workflow from another repo, won't do anything.

zhijie-yang:

---

## Description

This PR allows the reusable workflow to use the `.trivyignore` file from the rock's repo when the workflow is not run in the `canonical/oci-factory` repo.

### Related issues

Resolves #524 

### Test workflow run

https://github.com/canonical/golang-rock/actions/runs/16755185706

---

*Picture of a cool rock:*


.github/workflows/Test-Rock.yaml
```diff
@@ -245,6 +245,12 @@ jobs:
           repository: canonical/oci-factory
           ref: ${{ needs.configure-tests.outputs.oci-factory-ref }}
           fetch-depth: 1
+      
+      - name: Checkout Rock Repo
+        uses: actions/checkout@v4
+        if: github.repository != 'canonical/oci-factory'
+        with:
+          path: rock-repo
```
cjdcordeiro, line 253: this better be an env var since you're then using this path again in another step below

.github/workflows/Test-Rock.yaml
```diff
@@ -245,6 +245,12 @@ jobs:
           repository: canonical/oci-factory
           ref: ${{ needs.configure-tests.outputs.oci-factory-ref }}
           fetch-depth: 1
+      
+      - name: Checkout Rock Repo
+        uses: actions/checkout@v4
+        if: github.repository != 'canonical/oci-factory'
```
cjdcordeiro, line 251: since this is only needed for the repo's trivyignore path, should the condition also be `... && input.trivyignore-path`? (to avoid a checkout that's not gonna be used)

cjdcordeiro (changes requested):

.github/workflows/Test-Rock.yaml
zhijie-yang: Fixed in [56c57fe](https://github.com/canonical/oci-factory/pull/549/commits/56c57fe5cf23b7b27b51cb5375dcfcf3dfc7dcbc).

lczyk (approved):

PR #582, opened 2025-09-04
feat(node): update nodejs image

alesancor1:

---

## Description
Updating the nodejs image following the latest refactor of the repository. In terms of the rockcraft.yaml, this only changes the `deb-security-manifest` which is now using the one from https://github.com/canonical/rocks-security-manifest

---
<p align="center">
image
</p>

### DIFF
--- a/oci/node/documentation.yaml
+++ b/oci/node/documentation.yaml
@@ -23,5 +23,5 @@ debug:
     To get an interactive nodejs interpreter:
 
     ```bash
-    docker exec -it nodejs-container node
+    docker run --rm -it nodejs-container exec node
     ```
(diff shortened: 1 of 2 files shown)


oci/node/documentation.yaml
```diff
@@ -23,5 +23,5 @@ debug:
     To get an interactive nodejs interpreter:
 
     ```bash
-    docker exec -it nodejs-container node
+    docker run --rm -it nodejs-container exec node
```
zhijie-yang, line 26: I suspect this won't run correctly. `docker run` expects an image name instead of a container instance name. Should be either `docker run --rm -it ubuntu/node:18-24.04_edge`, or just keep the old command as it's creating a new interactive shell inside an already running container.

zhijie-yang (changes requested):

PR #619, opened 2025-09-18
feat(ci): use rebase in commit

zhijie-yang:

---

## Description

This PR intends to simplify the `commit-releases-json` action and mitigate the merge conflicts.

### DIFF
--- a/.github/actions/commit-releases-json/action.yaml
+++ b/.github/actions/commit-releases-json/action.yaml
@@ -49,6 +49,9 @@ runs:
         git add oci/${{ inputs.image-name }}/_releases.json
         git commit -m "${{ inputs.message }}"
         
-        git pull origin ${{ inputs.releases-branch }} --rebase
-
-        git push origin ${{ inputs.releases-branch }} $FORCE
+        for i in {1..10}; do
+          git pull origin ${{ inputs.releases-branch }} --rebase
+          git push origin ${{ inputs.releases-branch }} $FORCE && break
+          echo "Retrying push... attempt $i"
+          sleep 0.5
+        done


alesancor1:
I'm not sure on this. 

* What was the issue with the previous implementation, it seemed right to me, so why changing it?
* How will this mitigate the merge conflicts?

Imho if there are no real issues with the previous implementation I'd leave it as is, I believe it is easy to understand, it does not need to be a one liner.

Just being careful here, I'm not 100% confident about rebasing on CI/CD

zhijie-yang:
The previously used pattern follows [actions-x/commit](https://github.com/actions-x/commit). However, we do find many conflicts in previous CI runs, especially during weekly rebuilds (just taking 3 out of 10+ failed workflows, for example):

- https://github.com/canonical/oci-factory/actions/runs/17705241718/job/50316584939
- https://github.com/canonical/oci-factory/actions/runs/17705240231/job/50316518719
- https://github.com/canonical/oci-factory/actions/runs/17705232121/job/50316592800

I've reproduced the same scenarios locally, and `git pull --rebase` works quite fine, while the existing pattern errors out on merging.

In terms of rebase in CI/CD, I don't think it would be a problem, as we don't make any forced pushes, and there are also branch protection rules preventing us from committing non-linear commits into main.

lczyk:
a counterpoint from me: one liners can be a bit easier to parse than a block of code. if the rebase is the issue, we should specify `--rebase=false` to tell git to commit instead.

alesancor1:
I don't see how changing the merging strategy fixes this issue. Afaiu this issue is essentially a race condition, since it happens with weekly scheduled rebuilds, meaning that two or more runners are trying to push changes to the same remote. Even with the rebase strategy, I believe we would have something like:

```
Runner A:   C1 --- C2 --- C3  ----> origin/_releases
Runner B:   C1 --- C2 --- C3' ----> origin/_releases
```

Meaning we would get the same error, probably less frequently due to less steps happening between commit & push, but the error would still be there.

Assuming this is the case, the solution for me would be as simple as implementing a retry pattern, just `git pull` the changes before retrying the `git push` - whether you want to pass the `--rebase` flag to have a cleaner history instead of merge commits it's up to you.

alesancor1:
> If you look at the [log](https://github.com/canonical/oci-factory/actions/runs/17705241718/job/50316584939#step:12:64), `git` already identified the divergence of the branch. For our use case, pulling with `--rebase` is already sufficient.

This is a timing issue, in this case `git` identified the divergence due to `_released` being updated before `git merge` happened, but I don't think you can guarantee this will always be the case, especially as the amount of Rocks to rebuild keeps growing.

As I said, the `rebase` strategy will make it less likely for this to happen due to the time window for other workflows to introduce a conflict being reduced, but is not a final solution imo:

```
              Other workflow updates _releases
                             |
                             V
pull --rebase --------------------------> push ---> ERROR!
```

alesancor1:
Yes I get the point we are missing a `pull` to update the local state. My point is that we are not completely fixing the issue due to potential race condition. 

I think the fix for this could be simple enough to be in this PR, since the scope if "mitigate merge conflicts":

```bash
for i in {1..10}; do
  git pull origin ${{ inputs.releases-branch }} --rebase
  git push origin ${{ inputs.releases-branch }} $FORCE && break
  echo "Retrying push... attempt $i"
  sleep 0.5
done
```

Haven't tried it but something along those lines could work

zhijie-yang:
Hey @alesancor1, thanks for the approval. You can find the workflow run directly from the one triggered by the last push.

PR #628, opened 2025-09-24
feat: use self-hosted s390x runner

zhijie-yang:

---

## Description

This PR enables the OCI Factory to use the self-hosted s390x runner to build rocks, while we are having [an issue](https://bugs.launchpad.net/lpci/+bug/2125568) with the LPCI build.

Other changes:

- Bump the used runner for `Build-Rock` to noble
- Removed `actions/setup-python` from `Build-Rock` and its dependent custom action, since there aren't any available Python releases for s390x. Test workflow runs smoothly to have them removed.

### Related workflow runs:

- Failing workflow building `mysql` for s390x with LPCI: https://github.com/canonical/oci-factory/actions/runs/17977193659/job/51133574360
- Passing workflow building `mysql` for s390x with self-hosted s390x runner: https://github.com/canonical/oci-factory/actions/runs/17982280130

### DIFF
--- a/.github/workflows/Build-Rock.yaml
+++ b/.github/workflows/Build-Rock.yaml
@@ -90,9 +90,10 @@ jobs:
           submodules: "recursive"
           token: ${{ secrets.source-github-token }}
 
-      # Note: We skip setup-python here because there is no available Python release
-      # for architectures other than amd64 and arm64 from GitHub actions/setup-python.
-      # https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
+      - name: Installing Python 
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.x"
 
       - name: Installing Python requirements
         run: python3 -m pip install -r src/build_rock/configure/requirements.txt


.github/workflows/Build-Rock.yaml
```diff
@@ -182,16 +185,24 @@ jobs:
       - name: Building Target
         # TODO: Replace this retry action with bash equivalent for better testing
         uses: nick-fields/retry@v3.0.0
+        id: lpci-build
         with:
           timeout_minutes: 180
-          max_attempts: 4
+          max_attempts: 1
```
alesancor1, line 191: Why decreasing the attempts to only 1?

.github/workflows/Build-Rock.yaml
```diff
@@ -86,13 +90,12 @@ jobs:
           submodules: "recursive"
           token: ${{ secrets.source-github-token }}
 
-      - name: Installing Python 
-        uses: actions/setup-python@v5
-        with:
-          python-version: "3.x"
+      # Note: We skip setup-python here because there is no available Python release
+      # for architectures other than amd64 and arm64 from GitHub actions/setup-python.
+      # https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```
alesancor1, line 95: Afaiu this is running on `amd64` (the job has `runs-on: ubuntu-22.04`) so the setup python would still work no matter what.

alesancor1 (commented):
LGTM overall, just a couple of questions:

.github/workflows/Build-Rock.yaml
zhijie-yang: Oops, this is a leftover from when I was debugging the code. Fixed in 
3411b4f

.github/workflows/Build-Rock.yaml
```diff
@@ -92,7 +96,7 @@ jobs:
           python-version: "3.x"
 
       - name: Installing Python requirements
-        run: pip install -r src/build_rock/configure/requirements.txt
+        run: python3 -m pip install -r src/build_rock/configure/requirements.txt
```
alesancor1, line 99: I don't think this makes any difference but you can leave the command as it was as the setup-python action has been brought back

alesancor1 (approved):
LGTM, thanks!

PR #655, opened 2025-10-13
chore(ruby): promote to stable

cjdcordeiro:
Promote Ruby to `stable`.


alesancor1 (approved):
LGTM, though why are we manually doing `chisel cut` instead of using `stage-packages`?

Issue #642, opened 2025-10-01
[BUG] Current vulnerability check is giving false positives


simskij:
## Problem

The current vulnerability check utilized by the oci-factory is giving a lot of false positives, mainly as it does not check whether the code that contains the vulnerability is ever imported or called.

What happens at the moment is that we, every time a vulnerability is flagged, have to run `govulncheck` against the project to see whether the CVE is actually affecting us or not and add it to `trivyignore` if it isn't. A downside of this is that **if** the upstream ever starts to include the vulnerable code, it is already ignored and won't be flagged. 

Additionally, we have no way of flagging `trivyignores` for specific rock versions - only globally, which is a huge drawback.

## Suggested solution

This requires a two step solution:

- `trivyignore` files per rock version, or a way to pin ignores to version ranges. 
- When a go project is detected, `govulncheck` should be run, either as well or in place of trivy, and anything not present in both cve lists get excluded.

PR #672, opened 2025-10-31
feat(ci): build specific trivy ignore

zhijie-yang:

---

## Description

This PR introduces the build-specific (identified by the source-commit-directory triplet) vulnerability ignore mechanism by embedding the `ignored-vulnerabilities` field in the `image.yaml` file to allow a finer-grained vulnerability bypassing.

As a result, the schema version of the `image.yaml` is bumped to `v2`. The provided `ignored-vulnerabilities` information will be stored in the build metadata and automatically applied to the scheduled rebuilds.

A new schema will look like:

```
version: 2

upload:
  - source: "canonical/rocks-toolbox"
    commit: 17916dd5de270e61a6a3fd3f4661a6413a50fd6f
    directory: mock_rock/1.2
    release:
      1.2-22.04:
        end-of-life: "2030-05-01T00:00:00Z"
        risks:
          - beta
    ignored-vulnerabilities:
      - "CVE-2023-1234"
      - "CVE-2023-5678"
```

OCI Factory still works with the legacy `.trivyignore` files, but the users are encouraged to use the new schema to let OCI Factory reduce the false negative rates on vulnerability scans if a rock consists of multiple tracks.

The support of this feature in the CLI client will follow, after this PR is merged.

### Related issues

#642 

---

*Picture of a cool rock:*
image

### DIFF
--- a/src/image/prepare_single_image_build_matrix.py
+++ b/src/image/prepare_single_image_build_matrix.py
@@ -229,7 +229,7 @@ def find_eol_exceed_base_eol(builds: list[dict[str, Any]]):
     return tracks_eol_exceed_base_eol
 
 
-def flatten_ignored_vulnerabilities(builds: list[dict[str, Any]]):
+def flatten_ignored_vulnerabilities(builds: list[dict[str, Any]]) -> None:
     """Flatten ignored-vulnerabilities field for GH Actions matrix."""
     for build in builds:
         build["ignored-vulnerabilities"] = " ".join(


.github/workflows/Test-Rock.yaml
```diff
@@ -243,6 +246,12 @@ jobs:
     env:
       ROCK_REPO_DIR: "rock-repo"
     steps:
+      - name: Check mutual exclusivity of ignored vulnerabilities and trivyignore path
+        run: |
+          if [[ "${{ inputs.ignored-vulnerabilities }}" != "" && "${{ inputs.trivyignore-path }}" != "" ]]; then
+            echo "Error: 'ignored-vulnerabilities' and 'trivyignore-path' inputs are mutually exclusive."
+            exit 1
+          fi
```
alesancor1: Since `trivyignore-path` is now deprecated, a message should be added to encourage the use of `ignored-vulnerabilities` instead

zhijie-yang: I'm not sure whether we want to deprecate this right away. It takes time for the maintainers of the rocks to migrate to this new feature. I'd suggest we leave both supported for now, as the CI's behavior is designed to be compatible with both ways.

README.md
```diff
@@ -488,7 +489,8 @@ needed.
 |`test-efficiency`| False | bool | Enable Dive image efficiency test. Enabled by default. |
 |`test-vulnerabilities`| False | bool | Enable Trivy vulnerability test. Enabled by default. |
 |`vulnerability-report-artifact-name`| False | str | Custom filename for Trivy vulnerability report. |
-|`trivyignore-path`| False | str | Optional path to `.trivyignore` file used in vulnerability scan. |
+|`trivyignore-path`| False | str | Optional path to `.trivyignore` file used in vulnerability scan. When specifying this input, the `ignored-vulnerabilities` must be left empty. |
```
lczyk: > When specifying this input, the `ignored-vulnerabilities` must be left empty.

a test for this behavior woudl be great

README.md
```diff
@@ -204,6 +204,7 @@ Having said that this trigger's syntax is as follows:
 | upload[*].source | True | str | Git repository hosting the image's project. |
 | upload[*].commit | True | str | Specific reference in the source, where to run the build from. |
 | upload[*].directory | True | str | Path to the "rockcraft.yaml". Where the build will run from. |
+| upload[*].ignored-vulnerabilities | False | conlist[str, unique_items=true] | List of vulnerability IDs (CVE or GHSA) to ignore during vulnerability scanning. These IDs will be added to the image's build metadata for future reference. When specifying this field, the `.trivyignore` file will be ignored. |
```
lczyk: > When specifying this field, the `.trivyignore` file will be ignored.

same comment as here: https://github.com/canonical/oci-factory/pull/672/files#r2555680910

lczyk (changes requested):
woudl be great to get a tiny bit more testing for the case when someone specifies both `.trivyignore` and tries to use the `ignored-vulnerabilities` field. 1) to check the behavior there is correct and 2) to think through that code/behavior path and see where it fails. i expect using both `.trivyignore` and `ignored-vulnerabilities` to be a common mistake. worth adding a check which will warn/error a pipeline if both are specified?

README.md
zhijie-yang: Refer to this workflow run log: https://github.com/canonical/oci-factory/actions/runs/18971684730/job/54181217166#step:1:51

lczyk (approved):
re [this](https://github.com/canonical/oci-factory/pull/672#discussion_r2555680910) and [this](https://github.com/canonical/oci-factory/pull/672#discussion_r2555775054) comment: while a workflow run is a good proof that it works fine, and sufficient for an approve from me, note that there is a difference between an example of a successful run and a test. a successful run is a one-time proof. a test is also preventing regressions and documenting behavior _in perpetuum_.

README.md
zhijie-yang: Similarly, the logic written as an embedded script in the workflow does not support to be auto-tested.

Issue #718, opened 2025-12-19
[FEAT] allow usage of lfs artifacts in rock-craft


rpocase:
**Is your feature request related to a problem? Please describe.**
when reusing the oci-factory workflows i seem to be unable to use git-lfs objects within the build portion. i've included a snippet of the current implementation below. i've included the relevant workflow bits below. once i get to the build portion i see a checksum mismatch between the sha256sum specified in the repository.

```
name: Build and Publish Rocks to GHCR

on:
  push:
    branches: [main]
  pull_request:
    branches: [main]
  workflow_dispatch:

jobs:

  read-config:
    runs-on: ubuntu-latest
    outputs:
      ghcr-upload: ${{ steps.read-ci-config.outputs.ghcr-upload }}
      ghcr-cve-scan: ${{ steps.read-ci-config.outputs.ghcr-cve-scan }}
      build-matrix: ${{ steps.read-ci-config.outputs.build-matrix }}
      upload-matrix: ${{ steps.read-ci-config.outputs.upload-matrix }}
    steps:
      - name: Checkout repository
        uses: actions/checkout@v5
        with:
          lfs: true

      - name: Read .github/ci.yaml
        id: read-ci-config
        uses: canonical/rocks-template-actions/actions/read-ci-config@v1

  get-runners:
    runs-on: ubuntu-latest
    outputs:
      arch-map: ${{ steps.set-map.outputs.arch-map }}
    steps:
      - id: set-map
        run: |
          if [[ "${{ github.repository_owner }}" == "canonical" ]]; then
            echo 'arch-map={"amd64":["noble","X64","large"],"arm64":["noble","ARM64","large"]}' >> $GITHUB_OUTPUT
          else
            echo 'arch-map={"amd64":["ubuntu-24.04"],"arm64":["ubuntu-24.04-arm"]}' >> $GITHUB_OUTPUT
          fi

  build:
    needs: [get-runners, read-config]
    runs-on: ubuntu-latest
    strategy:
      matrix: ${{ fromJSON(needs.read-config.outputs.build-matrix) }}
    steps:
      - name: Checkout repository
        uses: actions/checkout@v5
        with:
          lfs: true
      - name: Build Rock
        uses: canonical/oci-factory/.github/workflows/Build-Rock.yaml@main
        with:
          rock-repo: ${{ github.repository }}
          rock-repo-commit: ${{ github.head_ref || github.ref_name }}
          rockfile-directory: ${{ matrix.directory }}
[...]

PR #725, opened 2026-01-13
feat: add lfs support to Build-Rock.yaml

rpocase:

@canonical/rocks 

---

## Description
This adds support for two new parameters:
* lfs: true or false. defaults to false
* lfs-pull-pattern: string pattern for which lfs files to pull. defaults to empty (signifying pull all)

these parameters are expected to be set in ci.yaml and flow through to the checkout for building the rock Build-Rock.yaml.

this has been tested in a private repo, but only for the single arch support variants. i excluded lcpi as i don't know if the builder supports extending the package set at run time - happy to add it if so

### Related issues
Fixes: #718 
Depends on: canonical/rocks-template-actions#14

---

cool-rock

### DIFF
--- a/.github/workflows/Build-Rock.yaml
+++ b/.github/workflows/Build-Rock.yaml
@@ -33,6 +33,10 @@ on:
         description: "Whether to download Git LFS files from target repo."
         type: boolean
         default: false
+      lfs-include:
+        description: "Value to provide to git lfs pull --include when downloading LFS files. If this is not provided then all LFS files are pulled"
+        type: string
+        default: ''
 
       # Parameters for multi-arch builds
       # specifying the runners with size label to prevent from running on private/tiobe runners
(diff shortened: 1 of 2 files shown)


.github/workflows/Build-Rock.yaml
```diff
@@ -184,6 +196,7 @@ jobs:
           submodules: "recursive"
           token: ${{ secrets.source-github-token }}
 
+
```
gjolly: This change is not needed

gjolly (commented):
LGTM just one nitpick.

.github/workflows/Build-Rock.yaml
```diff
@@ -29,6 +29,14 @@ on:
         description: "Directory in repository where to find the rockcraft.yaml file."
         type: string
         required: true
+      lfs:
+        description: "Whether to download Git LFS files from target repo."
+        type: boolean
+        default: false
+      lfs-include:
+        description: "Value to provide to git lfs pull --include when downloading LFS files."
```
alesancor1: Same here

lczyk (approved):

PR #767, opened 2026-02-13
Add mimir 2.7.14

sed-i:
- [ ] I have signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] I am following the [CONTRIBUTING](https://github.com/canonical/oci-factory/blob/main/CONTRIBUTING.md) guidelines

*Ping the @canonical/rocks team.*

---

## Description
This PR adds a false positive to trivyignore, superseding #763.

PR #769, opened 2026-02-13
Add CODEOWNERS file to o11y rocks

sed-i:
- [ ] I have signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] I am following the [CONTRIBUTING](https://github.com/canonical/oci-factory/blob/main/CONTRIBUTING.md) guidelines

---

## Description
This PR adds CODEOWNERS files to observability rocks.
This is helpful to avoid creating our own branches with new PRs for fixing rock builds. The most common scenario is adding a line to trivyignore.

### Related issues
Here are some examples to PRs I had to create because I couldn't push into existing PRs:
- #768
- #767
- #765

### DIFF
--- a/oci/alertmanager/CODEOWNERS
+++ b/oci/alertmanager/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @canonical/Observability
--- a/oci/blackbox-exporter/CODEOWNERS
+++ b/oci/blackbox-exporter/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @canonical/Observability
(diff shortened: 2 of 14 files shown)


zhijie-yang:
The easiest way for you would be onboarding your noctua bot to the OCI Factory, and use the CLI client to trigger the `Image` workflows, such that you don't even need to open a PR or make persist changes to the `image.yaml` file. We'll add support to specify `ignored-vulnerabilities` in the CLI client when triggering the workflow. With this, you can completely get rid of using PRs to build images.

The CLI client should have a easy integration to your existing workflows. I believe the IAM and data platform teams are already using this pattern.

PR #817, opened 2026-03-10
chore: update Trivy action version to 0.35.0

alesancor1:

---

## Description
Following the delete of many `trivy` releases due to a security incident (See related issue) - our action is now trying to pick up a release that no longer exists (unless cached in the runner).

Updating the action tag should fix this, as the latest action picks up `trivy-0.69.3` which does exists as a release.

### Related issues
* https://github.com/aquasecurity/trivy/discussions/10265

### DIFF
--- a/.github/actions/check-trivyignore/check-trivyignore-entries.sh
+++ b/.github/actions/check-trivyignore/check-trivyignore-entries.sh
@@ -10,11 +10,14 @@ trivyignores=$2
 
 temp_dir=$(mktemp -d)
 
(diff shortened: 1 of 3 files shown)


lczyk (approved):
looks good, although i do not like the fact that we cannot disambiguate between trivy failing and no vulnerabilities. since this is a reference check only this is an acceptable workaround, but i've made an issue to improve this at a later date:

https://github.com/canonical/oci-factory/issues/820

PR #851, opened 2026-03-30
docs: update charmed-opensearch documentation to v2

asanvaq:

---

## Description
Updated the documentation to v2.

### Related issues

---

*Picture of a cool rock:*

### DIFF
--- a/oci/charmed-opensearch/documentation.yaml
+++ b/oci/charmed-opensearch/documentation.yaml
@@ -1,9 +1,10 @@
-version: 1
+version: 2
 
 application: charmed-opensearch
-is_chiselled: False
 description: |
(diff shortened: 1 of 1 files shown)


asanvaq:
I get an error saying the docker repo does not exist, what do we do in those cases? Also, I am not 100% sure on the run cmd and the run conclusion, any advice is appreciated!

zhijie-yang:
I don't think this rock should still exist here, as both its repositories on ECR and DH have been removed.

@cjdcordeiro What happened to this rock? Should we drop it from OCI Factory?

cjdcordeiro:
I think it was one of the initial onboarding attempts, but beign a charm-specific rock, it didn't have a fit here.

zhijie-yang:
Dropped as the image is no longer maintained.

PR #889, opened 2026-04-10
chore: Add another litellm cve for airflow to trivyignore (patched in upstream)

shayancanonical:

---

## Description

The trivy scan of build for airflow-rock (3.1.8) found a new CVE: https://github.com/canonical/oci-factory/actions/runs/24136135570/attempts/1#summary-70427044805

The CVE stems from usage of litellm python package < 1.83.0
The upstream has patched to 1.83.0 (which is no affected by the vulnerability due to a fix), so this PR adds the discovered CVE to trivyignore to unblock build+publish of Airflwo image 3.1.8

### Related issues

---

*Picture of a cool rock:*

### DIFF
--- a/oci/airflow/.trivyignore
+++ b/oci/airflow/.trivyignore
@@ -212,11 +212,15 @@ CVE-2026-31958
 # This is a dependency for breeze (an Airflow dev tool) that is not included in the image
 CVE-2026-30922
 
-# (litellm): LiteLLM: Authentication bypass via OIDC userinfo cache key collision
 # Bump to patched version (1.83.0) merged upstream in `main`
 # Upstream PR: https://github.com/apache/airflow/pull/64701
+
+# (litellm): Authentication bypass via OIDC userinfo cache key collision
 CVE-2026-35030
+# (litellm): Privilege escalation via unrestricted proxy configuration endpoint
 CVE-2026-35029
+# (litellm): Password hash exposure and pass-the-hash authentication bypass
+GHSA-69x8-hrgq-fjj8
 
 # (tornado): Tornado has cookie attribute injection via .RequestHandler.set_cookie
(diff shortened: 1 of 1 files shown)


alesancor1:
I don't quite follow. If the upstream has patched the vulnerability, why is it being added to `.trivyignore` instead of bumping the version of the rock to match the upstream?

shayancanonical:
@alesancor1 I was under the impression that a vulnerability can be added to `.trivyignore` for the OCI factory image publish if:
- the vuln was reported upstream and there were some discussions around how the vuln would be addressed
- the vuln was patched upstream, but need not be backported to previous versions

The Airflow rock image currently build 3.1.8, which will have the vulnerability. Airflow has patched the package version that addresses the vulnerabilities in their development branch, which means that the next Airflow version will have the vulns addressed (they're not backporting to 3.1.8 as far as I understand)

This is why I proposed adding to `.trivyignore`, and adopting the patches by upgrading from `3.1.8` to the latest `3.1.x` or `3.2.x` when Airflow next releases

zhijie-yang (commented):
Since the upstream fix commit for the transitive vulnerable litellm missed the latest release, I'd suggest we give a green light to this one, as it only has one more day until EOL (as of this comment). I'd like to have it merged today, such that we don't have to update the image trigger again to set the EOL. wdyt @alesancor1 ?

P.S. @shayancanonical , we will deprecated the use of `.trivyignore` files very shortly, so please use the v2 trigger and embed the ignored vulnerabilities in it from the next PR.

alesancor1 (approved):

alesancor1:
Re: `v2` image trigger, see the [go image](https://github.com/canonical/oci-factory/blob/7b181f1b01220af97348d66dde106efa9af83cbe/oci/go/image.yaml#L15) for reference

PR #924, opened 2026-04-17
docs: update to v2 php

asanvaq:

---

## Description
Update docs to v2

### DIFF
--- a/oci/php/documentation.yaml
+++ b/oci/php/documentation.yaml
@@ -2,16 +2,30 @@ version: 2
 
 application: PHP
 description: |
+  PHP is a general-purpose scripting language well-suited for Web
+  development. Fast, flexible and pragmatic, PHP powers everything
+  from your blog to the most popular websites in the world.
 
-  PHP is a general-purpose scripting language well-suited for web development.
-  Fast, flexible, and pragmatic, PHP powers everything from your blog to the most popular websites in the world.
 website: https://www.php.net/
 issues: https://github.com/canonical/php-rock/issues
 source-code: https://github.com/canonical/php-rock
 
 docker:
+  run_cmd: exec php <args>
   run_conclusion: |
-    The container starts in detached mode.
+    PHP will execute any given instruction, for example running php code from input:
+
+    ```bash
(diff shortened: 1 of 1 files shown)


oci/php/documentation.yaml
```diff
@@ -1,33 +1,18 @@
-version: 1
+version: 2
 
 application: PHP
-is_chiselled: true
 description: |
-  PHP is a general-purpose scripting language well-suited for Web
-  development. Fast, flexible and pragmatic, PHP powers everything
-  from your blog to the most popular websites in the world.
+  PHP is a rock for PHP maintained by [Canonical](https://canonical.com).
```
zhijie-yang:
```suggestion
```

This is redundant, as it is automatically included by the template.

oci/php/documentation.yaml
```diff
@@ -1,33 +1,18 @@
...
-
-    ```bash
-    docker run --rm ubuntu/php:8.3-24.04_edge exec php -r "echo 'hello, world! In PHP!';"
-    ```
-
-    To run a php server for development purposes:
-
-    ```bash
-    docker run -d --rm --name php-container -p 8080:8080 -v /my-app:/app ubuntu/php:8.3-24.04_edge exec php -S 0.0.0.0:8080 -t /app
-    ```
```
zhijie-yang, line 24: Why don't we use the server run as the docker instructions?

@alesancor1 will it be okay to just include a plain `docker run` command in the doc?

zhijie-yang (changes requested):

oci/php/documentation.yaml
alesancor1: The server is just one of many available options for the PHP CLI, I've modified the docker run section to present the usage via `exec php <args>` while adding the previous usage examples in the `run_conclusion` section.

See https://github.com/canonical/oci-factory/pull/924/commits/a2d4c7f5681f447d6374d5971329256ffef70731

Check the rendered markdown [here](https://github.com/canonical/oci-factory/actions/runs/24831430064?pr=924)

zhijie-yang (approved):
Thanks for the update. This matches what we have in v1.

PR #931, opened 2026-04-17
docs: migrate to v2 zookeeper

asanvaq:

---

## Description
Migrate docs to v2.

### DIFF
--- a/oci/zookeeper/documentation.yaml
+++ b/oci/zookeeper/documentation.yaml
@@ -1,59 +1,38 @@
-version: 1
+version: 2
 application: zookeeper
-is_chiselled: False
 description: |
-  Current Apache ZooKeeper Docker Image from Canonical, based on Ubuntu. 
-  Receives security updates and tracks the newest combination of Apache 
-  ZooKeeper and Ubuntu. **This repository is free to use and exempted from 
-  per-user rate limits.**
-  
-  # About Apache ZooKeeper
-  ZooKeeper is a centralized service for maintaining configuration information, 
-  naming, providing distributed synchronization, and providing group services. 
-  All of these kinds of services are used in some form or another by 
-  distributed applications. Read more on the Apache website. This image is 
-  shipped in support of the Apache Kafka image.
-  
-  Please note that the images tagged up to 3.1 are Dockerfile-base images,
-  whereas from version 3.8.2 onward the images are now rocks. As such the
-  entrypoint is now Pebble. Read more on the
-  [Rockcraft docs](https://canonical-rockcraft.readthedocs-hosted.com/en/latest/explanation/rocks/).
+  ZooKeeper is a centralized service for maintaining configuration information,
+  naming, providing distributed synchronization, and providing group services.
+  All of these kinds of services are used in some form or another by
+  distributed applications. This rock is shipped in support of the Apache Kafka rock.
 
-# --- USAGE INFORMATION ---
+  Please note that rocks tagged up to 3.1 are Dockerfile-based images,
+  whereas from version 3.8.2 onward the rocks are chiseled rocks.
+website: https://zookeeper.apache.org/
+issues: https://github.com/canonical/zookeeper-rock/issues
+source-code: https://github.com/canonical/zookeeper-rock
+
 docker:
-  parameters:
-    - -p 2181:2181
-  access: Access your ZooKeeper server at `http://localhost:2181`.
-parameters:
-  - type: -e
-    value: 'TZ=UTC'
+  parameters: >-
+    -p 2181:2181
+  run_conclusion: |
+    ZooKeeper starts and serves client traffic on port 2181.
+    Access your ZooKeeper server at localhost:2181 after running with a port mapping.
+
+config:
+  TZ:
+    type: env
     description: Timezone.
-  - type: -p
-    value: '2181:2181'
-    description: Expose ZooKeeper server on `localhost:2181`.
-  - type: -v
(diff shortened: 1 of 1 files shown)


oci/zookeeper/documentation.yaml
```diff
@@ -1,59 +1,38 @@
...
-  whereas from version 3.8.2 onward the images are now rocks. As such the
-  entrypoint is now Pebble. Read more on the
-  [Rockcraft docs](https://canonical-rockcraft.readthedocs-hosted.com/en/latest/explanation/rocks/).
+  ZooKeeper is a centralized service for maintaining configuration information,
+  naming, providing distributed synchronization, and providing group services.
+  All of these kinds of services are used in some form or another by
+  distributed applications. This rock is shipped in support of the Apache Kafka rock.
+
+  Please note that rocks tagged up to 3.1 are Dockerfile-based images,
+  whereas from version 3.8.2 onward the rocks are chiseled rocks.
```
alesancor1, line 10: We can remove this bit, but being this the case we need to make use of `docker.legacy` fields to ensure it is being covered by the docs.

Could you provide some information about how to use the legacy docker images? @deusebio @imanenami 

Check [this section](https://github.com/canonical/oci-factory#documentation-trigger-file) of the docs for reference about the `docker.legacy` field.

oci/zookeeper/documentation.yaml
```diff
@@ -1,59 +1,38 @@
...
-    ```
+  '`-v <volume>:/var/lib/zookeeper/data`':
+    type: mount
+    description: Persist data in a Docker volume named zookeeperData. Keep the mount point consistent with the configuration property dataDir.
+  '`-v <volume>:/var/lib/zookeeper/data-log`':
+    type: mount
+    description: Persist data in a Docker volume named zookeeperLogData. Keep the mount point consistent with the configuration property dataLogDir.
+  '`-p <port>:2181`':
+    type: port
+    description: Expose port 2181 on the local host.
```
alesancor1:
```suggestion
    description: Expose the container's port `2181` on the host's `<port>`.
```

oci/zookeeper/documentation.yaml
```diff
@@ -1,59 +1,38 @@
...
-    - -p 2181:2181
-  access: Access your ZooKeeper server at `http://localhost:2181`.
-parameters:
-  - type: -e
-    value: 'TZ=UTC'
+  parameters: >-
+    -p 2181:2181
+  run_conclusion: |
+    ZooKeeper starts and serves client traffic on port 2181.
+    Access your ZooKeeper server at localhost:2181 after running with a port mapping.
```
alesancor1:
```suggestion
    Access the ZooKeeper server at `localhost:2181`.
```

alesancor1 (changes requested):
We need to document the previous behaviour using `docker.legacy` fields.

oci/zookeeper/documentation.yaml
```diff
@@ -1,59 +1,37 @@
...
-    To get an interactive shell:
-    ```bash
-    docker exec -it <zookeeper-container> /bin/bash
-    ```
+  '`-v <volume>:/var/lib/zookeeper/data`':
+    type: mount
+    description: Persist data in a Docker volume named zookeeperData. Keep the mount point consistent with the configuration property dataDir.
+  '`-v <volume>:/var/lib/zookeeper/data-log`':
+    type: mount
+    description: Persist data in a Docker volume named zookeeperLogData. Keep the mount point consistent with the configuration property dataLogDir.
```
zhijie-yang:
```suggestion
    description: Persist data mounted inside the container. The mount point should be consistent with the configuration property `dataLogDir`.
```

oci/zookeeper/documentation.yaml
```diff
@@ -1,59 +1,37 @@
...
-    docker logs -f <zookeeper-container>
-    ```
-    
-    To get an interactive shell:
-    ```bash
-    docker exec -it <zookeeper-container> /bin/bash
-    ```
+  '`-v <volume>:/var/lib/zookeeper/data`':
+    type: mount
+    description: Persist data in a Docker volume named zookeeperData. Keep the mount point consistent with the configuration property dataDir.
```
zhijie-yang:
```suggestion
    description: Persist data mounted into the container. The mount point should be consistent with the configuration property `dataDir`.
```

zhijie-yang (changes requested):

oci/zookeeper/documentation.yaml
```diff
@@ -1,59 +1,37 @@
...
-  parameters:
-    - -p 2181:2181
-  access: Access your ZooKeeper server at `http://localhost:2181`.
-parameters:
-  - type: -e
-    value: 'TZ=UTC'
+  parameters: >-
+    -p 2181:2181
+  run_conclusion: |
+    Access the ZooKeeper server at `localhost:2181`.
```
zhijie-yang, line 19:
```suggestion
  parameters: >-
    -p 2181:2181
  run_conclusion: |
    Access the ZooKeeper server at `localhost:2181`.
  legacy:
    parameters: >-
      -p 2181:2181
    run_conclusion: |
      Access the ZooKeeper server at `localhost:2181`.
```

I believe this should be good. At least the image runs normally and functioning with `docker run --rm -p 2181:2181 ubuntu/zookeeper:3.1-22.04_edge`

zhijie-yang (approved):
Looks good to me. Although this is not yet "perfect", at least we have an equivalent doc compared to the v1 schema, and nothing from the old doc is undermined in this migration.

PR #943, opened 2026-04-17
feat: add curl rock

cjdcordeiro:

---

## Description

Onboard the `curl` rock

### DIFF
--- a/oci/curl/image.yaml
+++ b/oci/curl/image.yaml
@@ -1,4 +1,4 @@
-version: 1
+version: 2
 
 upload:
   - source: canonical/curl-rock
--- a/oci/curl/image.yaml
+++ b/oci/curl/image.yaml
@@ -1,0 +1,11 @@
+version: 1
+
+upload:
+  - source: canonical/curl-rock
+    commit: 61187f2180b06f8525ffdc53d5e3beb63d83cc23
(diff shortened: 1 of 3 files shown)


oci/curl/image.yaml
```diff
@@ -0,0 +1,11 @@
+version: 1
```
alesancor1: Let's try to use the `v2` of the image trigger from now on.

alesancor1 (approved):
LGTM, thanks!

oci/curl/image.yaml
cjdcordeiro: done. why was `v2` introduced though? iirc the latest changes didn't create any regressions or disruptive patterns

alesancor1: introduces `ignored-vulnerabilities` field to get rid of all the `.trivyignore` files

Not really required for this one, since there are not any CVEs (yet), hence why I approved.

cjdcordeiro: for the future, it think it is a better practice to simply allow optional fields to simply come along within the same version. If they don't create disruption, but rather augment the schema with optional capabilities, then we might want to avoid creating a new version

zhijie-yang (approved):

PR #961, opened 2026-04-23
fix: use --latest when --devel is outdated

alesancor1:

---

## Description
Since 26.04 is released,  there's nothing in `devel` until `26.10` is opened. So to prevent build failures, we use `--latest` in this situation, since is where `devel` should point to.

### DIFF
--- a/src/uploads/infer_image_track.py
+++ b/src/uploads/infer_image_track.py
@@ -40,7 +40,7 @@ def get_base_and_track(rockcraft_yaml) -> tuple[str, str]:
                     "devel is outdated. Using the latest release instead."
                 )
                 distro_info = subprocess.check_output(
-                    ["ubuntu-distro-info", "--latest", "--codename"], universal_newlines=True
+                    ["ubuntu-distro-info", "--latest", "--release"], universal_newlines=True
                 )
             else:
                 raise e
--- a/src/uploads/infer_image_track.py
+++ b/src/uploads/infer_image_track.py
@@ -32,7 +32,9 @@ def get_base_and_track(rockcraft_yaml) -> tuple[str, str]:
     if rock_base == "devel":
         try:
             distro_info = subprocess.check_output(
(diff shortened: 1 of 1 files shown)


src/uploads/infer_image_track.py
```diff
@@ -30,9 +30,23 @@ def get_base_and_track(rockcraft_yaml) -> tuple[str, str]:
...
 
     if rock_base == "devel":
-        rock_base = subprocess.check_output(
-            ["ubuntu-distro-info", "--devel", "--release"], universal_newlines=True
-        ).split()[0]
+        try:
+            distro_info = subprocess.check_output(
+                ["ubuntu-distro-info", "--devel", "--release"],
+                universal_newlines=True,
+                stderr=subprocess.STDOUT,
```
zhijie-yang, line 37: I'd use a `capture_output=True` here instead of redirecting the stderr to stdout, since you'll be parsing the stdout later on.

zhijie-yang (approved):
A minor comment here. Good for a fix.

src/uploads/infer_image_track.py
alesancor1: `capture_output` belongs to `subprocess.run` so we'd need to change the logic to something like:

```python
result = subprocess.run(
    ["ubuntu-distro-info", "--devel", "--release"],
    capture_output=True,
    text=True,
)

if result.returncode == 0:
    distro_info = result.stdout
elif result.returncode == 1 and "Distribution data outdated" in result.stderr:
    result = subprocess.run(
        ["ubuntu-distro-info", "--latest", "--release"],
        capture_output=True,
        text=True,
        check=True,
    )
    distro_info = result.stdout
else:
    result.check_returncode()  # raise the error

rock_base = distro_info.split()[0]
```

Something along those lines, I don't have a strong opinion on whether to use one or the other tbh, I'm happy to switch if you think `subprocess.run` is more appropiate here

zhijie-yang: happy to keep the current impl.

PR #962, opened 2026-04-23
chore(ci): install doc deps when testing

zhijie-yang:

---

## Description

Resolves wf slient error: https://github.com/canonical/oci-factory/actions/runs/24828239848/job/72681217736?pr=961

### Related issues

---

*Picture of a cool rock:*

### DIFF
--- a/.github/workflows/_Test-OCI-Factory.yaml
+++ b/.github/workflows/_Test-OCI-Factory.yaml
@@ -57,6 +57,9 @@ jobs:
       # build_rock/configure requirements
       - run: pip install -r src/build_rock/configure/requirements.txt
 
+      # docs requirements
+      - run: pip install -r src/docs/requirements.txt
+
       - name: Run pytest
         continue-on-error: true
         run: |


.github/workflows/_Test-OCI-Factory.yaml
alesancor1, line 79: Sort of related to silent errors:

Using `continue-on-error: true` will cause the test to be green even if it fails, which IMO can cause confusion and lead to some issues being overlooked. 

I'd rather remove `continue-on-error` and run the summary and the result upload with `always()`

```suggestion
- name: Run pytest
run: |
         python3 -m pytest --junit-xml "${{ env.PYTEST_RESULT_PATH }}"

     - name: Generate Summary
if: always()
run: |
         python3 -m tools.junit_to_markdown --input-junit "${{ env.PYTEST_RESULT_PATH }}" >> $GITHUB_STEP_SUMMARY

     - name: Upload pytest Result
if: always()
uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f
with:
name: ${{ env.PYTEST_RESULT_PATH }}
path: ${{ env.PYTEST_RESULT_PATH }}
if-no-files-found: error
```

This way we'll see the red dot in the PRs directly.

alesancor1 (approved):
Good catch, LGTM!

Leaving a comment below which  is related to silent failures. Not to be addressed here necessarily. Feel free to open a new PR or reject the change straightaway.

lczyk (approved):

.github/workflows/_Test-OCI-Factory.yaml
zhijie-yang: Thanks for the suggest. Will put this into another PR.

PR #1020, opened 2026-05-21
fix(ci): grant proper write permissions in workflow

zhijie-yang:

---

## Description

The new organizational complice policy enforces workflow permissions to be read-only by default. This PR gives the workflows the correct permissions to write to packages.

### Related issues

N/A, but failing workflows.
---

*Picture of a cool rock:*

### DIFF
--- a/.github/workflows/Announcements.yaml
+++ b/.github/workflows/Announcements.yaml
@@ -14,8 +14,6 @@ jobs:
   get-contacts:
     name: Get contacts and image info
     runs-on: ubuntu-22.04
-    permissions:
-      actions: write
     outputs:
       mattermost-channels: ${{ steps.get-contacts.outputs.mattermost-channels }}
       oci-img-name: ${{ steps.get-image-name.outputs.img-name }}
--- a/.github/workflows/Announcements.yaml
+++ b/.github/workflows/Announcements.yaml
@@ -14,6 +14,8 @@ jobs:
   get-contacts:
     name: Get contacts and image info
     runs-on: ubuntu-22.04
+    permissions:
+      actions: write
     outputs:
       mattermost-channels: ${{ steps.get-contacts.outputs.mattermost-channels }}
       oci-img-name: ${{ steps.get-image-name.outputs.img-name }}
--- a/.github/workflows/Documentation.yaml
+++ b/.github/workflows/Documentation.yaml
@@ -32,8 +32,6 @@ jobs:
   validate-documentation-request:
     runs-on: ubuntu-22.04
     name: Validate documentation request
-    permissions:
-      actions: write
     outputs:
       oci-img-path: ${{ steps.validate-image.outputs.img-path }}
       oci-img-name: ${{ steps.validate-image.outputs.img-name }}
--- a/.github/workflows/Documentation.yaml
+++ b/.github/workflows/Documentation.yaml
@@ -32,6 +32,8 @@ jobs:
   validate-documentation-request:
     runs-on: ubuntu-22.04
     name: Validate documentation request
+    permissions:
+      actions: write
     outputs:
       oci-img-path: ${{ steps.validate-image.outputs.img-path }}
       oci-img-name: ${{ steps.validate-image.outputs.img-name }}
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -294,6 +294,8 @@ jobs:
     
     permissions:
       packages: write
+      # Used by step "Create Git tag"
+      contents: write
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
 
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -290,10 +290,10 @@ jobs:
     outputs:
       artefacts-hashes: ${{ steps.artefacts-hashes.outputs.hashes }}
     
-    permissions:
-      packages: write
-      # Used by step "Create Git tag"
-      contents: write
+    # permissions:
+    #   packages: write
+    #   # Used by step "Create Git tag"
+    #   contents: write
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
 
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -49,6 +49,8 @@ jobs:
   prepare-build:
     runs-on: ubuntu-22.04
     name: Prepare build
+    permissions:
+      actions: write
     outputs:
       build-matrix: ${{ steps.prepare-matrix.outputs.build-matrix }}
       release-to: ${{ steps.prepare-matrix.outputs.release-to }}
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -563,6 +563,9 @@ jobs:
   release:
     name: Release
     needs: [prepare-build, prepare-releases]
+    permissions:
+      actions: write
+      contents: write
     uses: ./.github/workflows/Release.yaml
     with:
       oci-image-name: "${{ needs.prepare-build.outputs.oci-img-name }}"
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -49,8 +49,6 @@ jobs:
   prepare-build:
     runs-on: ubuntu-22.04
     name: Prepare build
-    permissions:
-      actions: write
     outputs:
       build-matrix: ${{ steps.prepare-matrix.outputs.build-matrix }}
       release-to: ${{ steps.prepare-matrix.outputs.release-to }}
@@ -566,7 +564,6 @@ jobs:
     name: Release
     needs: [prepare-build, prepare-releases]
     permissions:
-      actions: write
       contents: write
     uses: ./.github/workflows/Release.yaml
     with:
--- a/.github/workflows/Image.yaml
+++ b/.github/workflows/Image.yaml
@@ -290,10 +290,10 @@ jobs:
     outputs:
       artefacts-hashes: ${{ steps.artefacts-hashes.outputs.hashes }}
     
-    # permissions:
-    #   packages: write
-    #   # Used by step "Create Git tag"
-    #   contents: write
+    permissions:
+      packages: write
+      # Used by step "Create Git tag"
+      contents: write
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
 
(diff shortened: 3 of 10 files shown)


.github/workflows/_Test-OCI-Factory.yaml
```diff
@@ -139,3 +141,7 @@ jobs:
       oci-image-name: "mock-rock"
       upload: true
     secrets: inherit
+    permissions:
+      packages: write
+      actions: write
+      contents: write
```
alesancor1, line 144: I'd avoid granting permissions here at the top level. The `image.yaml` workflow is the one that should handle this permissions within the correct job (or if the job uses another workflow, then inside that workflow).

TL;DR always try to define the permissions to the narrowest scope :+1:

.github/workflows/_Test-OCI-Factory.yaml
```diff
@@ -28,6 +28,8 @@ jobs:
   access-check:
     name: Validate access to mock-rock
     runs-on: ubuntu-latest
+    permissions:
+      actions: write
```
alesancor1: Why? According to [the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions) this permission would allow to manage workflows (trigger / cancel, etc.) I don't recall this being needed in OCI-Factory

.github/workflows/Announcements.yaml
```diff
@@ -14,6 +14,8 @@ jobs:
   get-contacts:
     name: Get contacts and image info
     runs-on: ubuntu-22.04
+    permissions:
+      actions: write
```
alesancor1: Same here, does this job require to do any write operations to the actions in the repo?

.github/workflows/Image.yaml
```diff
@@ -558,6 +565,9 @@ jobs:
   release:
     name: Release
     needs: [prepare-build, prepare-releases]
+    permissions:
+      actions: write
+      contents: write
```
alesancor1: The `Release.yaml` workflow should be the one defining this

alesancor1 (changes requested):
I have some concerns with the scope at which the permissions are being granted, let's try to narrow it down to ensure that no job has more permissions that it needs.

Also I'm unsure about the `actions: write` permission, IIRC we don't really trigger or stop other workflows within our actions, do we?

See comments below

.github/workflows/_Test-OCI-Factory.yaml
zhijie-yang: These is no way not to put it here. Workflow becomes invalid without this. https://github.com/canonical/oci-factory/actions/runs/26219360359

.github/workflows/_Test-OCI-Factory.yaml
zhijie-yang: `validate-actor` action won't be able to cancel the workflow runs without this.

.github/workflows/Announcements.yaml
zhijie-yang: Same reason as above.

.github/workflows/Image.yaml
zhijie-yang: `Image.yaml` workflow has to have this since it's the caller. Otherwise, the workflow becomes invalid.

.github/workflows/_Test-OCI-Factory.yaml
lczyk: yeah. alas, when you use a reusable workflow, it must not have greater permissions than you (aka the caller). prev this worked because GITHUB_TOKEN had higher default perms, but now it would error. i've seen the same thing in chisel-releases.

lczyk: i don't think this is actually adding more permissions to actions than before, its jut being more explicit about them. what i *would* suggest is for us to be even more explicit actually, and next to every `permissions:` write justification e.g. 

```yaml
    # validate-actor needs `actions: write` to cancel jobs. we as a caller of the workflow must therefore have those permissions
    permissions:
      actions: write
```

lczyk: this weirdly makes me think of rust's ownership model actually... 🤔 a lot of headaches in rust are alleviated by, as opposed to mutating something in the called function, just returning an immutable struct with to the caller with the instructions as to what ought to happen. the caller then never gives up the ownership but passes a read-only reference. to put it into the more precient language of actions: what is validate-actor does cannot cancel jobs? instead, what if it only returns an output to tell the caller workflow (aka here) that it wants to cancel the job, and then the write permissions can be scoped to one, easily auditable place as opposed to threaded through the entire workflow. 🤔 

edit: quickly scoped it in https://github.com/canonical/oci-factory/pull/1023 but the permissions are set per-job, not per-step so we'd need to now split things into multiple jobs and this kind of refactor would then blow up v quickly. idea is sound in principle, i think, but not really feasible, certainly not as a small piece of work. might be worth keeping in mind as a sec-oriented design pattern for actions in the future.

lczyk (changes requested):

.github/workflows/_Test-OCI-Factory.yaml
cjdcordeiro: but that step is using a fine grained token. So why aren't that token's permissions enough?

cjdcordeiro: if anything, I'd even consider not cancelling the workflow. i.e. if the step fails, the remaining jobs shouldn't run anyway

.github/workflows/Documentation.yaml
```diff
@@ -32,6 +32,8 @@ jobs:
   validate-documentation-request:
     runs-on: ubuntu-22.04
     name: Validate documentation request
+    permissions:
+      actions: write
```
cjdcordeiro: it seems like a lot of these are because of the `validate-actor` action. That raises the question on whether it should be doing what it is doing atm

.github/workflows/Release.yaml
cjdcordeiro, line 247: for future reference, we should create the tag and release ourselves. no need to call 3rd party actions

cjdcordeiro (changes requested):
please add comments on every permission setting, to track why said permissions are being granted. 

Also, potentially out of scope for this PR, but absolutely needed as a part of the due diligence: for every job that has `contents: write`, check if the job can be further split into smaller jobs, such that the powerful GITHUB_TOKEN isn't lingering around for grabs by 3rd party actions that could be in another unprivileged job

.github/workflows/Documentation.yaml
zhijie-yang: For what `validate-actor` action does, it does fine on gating the workflows, given that we have many workflows triggered by bots (e.g., `[canonical-data-platform-bot](https://github.com/canonical-data-platform-bot)`)

.github/workflows/_Test-OCI-Factory.yaml
zhijie-yang: > but that step is using a fine grained token. So why aren't that token's permissions enough?

Yes, with [the fix]([c260425](https://github.com/canonical/oci-factory/pull/1020/commits/c26042566c91eb207e6814f73e7805a29f836f98)), it is sufficient.

> if anything, I'd even consider not cancelling the workflow. i.e. if the step fails, the remaining jobs shouldn't run anyway

No. We discussed this two years ago. The idea behind cancel is "not letting the workflow with `if: !canceled()` run." The workflow should halt and exit immediately if gated.

.github/workflows/Documentation.yaml
zhijie-yang: fixed and removed this permission.

.github/workflows/Image.yaml
zhijie-yang: `actions: write` removed. `contents: write` has to stay.

.github/workflows/Announcements.yaml
zhijie-yang: removed.

.github/workflows/_Test-OCI-Factory.yaml
alesancor1: Re-opening this.

I see that `actions: write` is gone since the `validate-actor` now runs with the rocksbot token, so that fixes the worklow becoming invalid.

I'm concerned about `packages: write` and `contents: write`, especially considering this workflow is being called on PRs. They should not be granted here.

zhijie-yang: They MUST be granted here, as this job triggers a FULLY CAPABLE Image workflow.

zhijie-yang: There is no way to call a fully capable Image workflow without granting these permissions. Without applying a global permission setting in the Image workflow, these granted permissions in the caller workflow won't necessarily expand the scope.

zhijie-yang: Unless you add a `workflow-run` trigger to the Image workflow, but these two permissions will still be required from the Image workflow and therefore make no difference.

alesancor1: Yep, reading deeper into the docs seems that the callee is constrained to the permissions from the caller. But this is still concerning...

Considering the amount of CI/CD actions in this project, what I'd suggest is to make the permissions explicit for every job. Also because the job permissions override the workflow's, so even if you set `contents: write` here you can still set `contents: read` in the callee jobs, and leave `write` only where it needs it.

zhijie-yang: There is no need to do so. Refer to the [doc](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#supported-keywords-for-jobs-that-call-a-reusable-workflow), the first point in the "Notes" clearly states that the default permission will apply if no explicit permissions are given.

alesancor1: Thanks for the ref to the docs, that clarifies it :+1:

zhijie-yang:
As discussed offline, we won't impose explicit `permission.*.read` in this PR, as the default permissions are already `read`, and jobs/workflows without explicit permission granting will stick to the default permissions[1].

[1] First point in "Note": https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#supported-keywords-for-jobs-that-call-a-reusable-workflow

alesancor1 (approved):
LGTM, thanks!

cjdcordeiro (commented):
pls address/comment/resolve the previous comments before re-requesting review

.github/workflows/Release.yaml
zhijie-yang: issue created as https://github.com/canonical/oci-factory/issues/1024

zhijie-yang:
all closed.

PR #1021, opened 2026-05-21
feat(sloth): create .trivyignore to ignore certain CVEs

lucabello:

---

## Description
Ignoring some CVEs that are not affecting the sloth code as per `govulncheck` scan.

---

*Picture of a cool rock:* 🗿

### DIFF
--- a/oci/sloth/.trivyignore
+++ b/oci/sloth/.trivyignore
@@ -1,0 +1,8 @@
+# Upstream CVEs
+
+## Upstream not affected
+# github.com/prometheus/prometheus
+CVE-2026-42151
+CVE-2026-42154
+# go.opentelemetry.io/otel
+CVE-2026-29181


alesancor1:
Hey @lucabello thanks for the PR!

Actually, the preferred way currently is to use the `v2` image trigger to define the ignored CVEs using the `ignored-vulnerabilities` field in the yaml. See the docs in the [README](https://github.com/canonical/oci-factory#image-trigger-file).


For reference, you can check the loki rock:

https://github.com/canonical/oci-factory/blob/b35f55ab3dbac8ca7b4d3b86368522d0880d78eb/oci/loki/image.yaml#L1-L25

Note the `version: 2` at the top!

lucabello:
@alesancor1 what about ignoring vulenrabilities for images that are already in oci-factory? Should we re-add them via the `image.yaml` file and put the vulnerabilities there?

alesancor1:
@lucabello OCI-Factory still supports the `v1` trigger, so there's no rush of migrating all images right now. When CVE is found in a published image then you can make use of the `v2` instead of the `.trivyignore` file.

zhijie-yang (changes requested):
Thanks for this PR. As of the time this PR is authored (21 May), there are already no in-release tracks from this rock. (see [_releases.json](https://github.com/canonical/oci-factory/blob/_releases/oci/sloth/_releases.json)). How would you proceed with this rock?

As all the tracks are EOL'ed, if you want to revive this rock, you'd need a new image trigger to bump the EOL dates, which would be a nice chance for you to start with the v2 trigger.

On the other hand,

>  what about ignoring vulenrabilities for images that are already in oci-factory? Should we re-add them via the `image.yaml` file and put the vulnerabilities there?

This case for `sloth` does not hold, as all the tracks are EOL, and there won't be any vulnerability scans for them. You can just include everything in the updated `image.yaml` file (i.e., the new tracks/builds you want to build + the old tracks/builds you want to maintain + the per-build vulnerability ignore list for them).

Issue #1024, opened 2026-05-21
[FEAT] replace the third-party actions with `gh` calls


zhijie-yang:
The use of third-party actions should be replaced by `gh` calls in Image workflow, e.g. https://github.com/canonical/oci-factory/pull/1020#discussion_r3281555406

PR #1025, opened 2026-05-22
ci: replace third-party tag/release actions with gh calls

lczyk:

---

## Description

attempt at replacing `rickstaa/action-create-tag` and `softprops/action-gh-release` with `gh` equivalents.

### Related issues

closes(?) #1024

### DIFF
--- a/.github/workflows/Release.yaml
+++ b/.github/workflows/Release.yaml
@@ -259,10 +259,17 @@ jobs:
             -f "tagger[email]=${GITHUB_ACTOR}@users.noreply.github.com" \
             -f "tagger[date]=$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
             --jq .sha)
-          # Create the tag ref (previous step deleted any existing tag via --cleanup-tag)
-          gh api "repos/${REPO}/git/refs" \
-            -f ref="refs/tags/${TAG}" \
-            -f sha="${tag_sha}"
+          # Preserve the prior force-push-tag behaviour: update the ref when
+          # it already exists, otherwise create it.
+          if gh api "repos/${REPO}/git/ref/tags/${TAG}" >/dev/null 2>&1; then
+            gh api --method PATCH "repos/${REPO}/git/refs/tags/${TAG}" \
+              -f sha="${tag_sha}" \
+              -F force=true
+          else
+            gh api "repos/${REPO}/git/refs" \
+              -f ref="refs/tags/${TAG}" \
+              -f sha="${tag_sha}"
+          fi
 
       - name: Create GitHub release
         env:
--- a/.github/workflows/Release.yaml
+++ b/.github/workflows/Release.yaml
@@ -245,31 +245,34 @@ jobs:
           if [[ "$RUNNER_DEBUG" == "1" ]]; then
             set -x
           fi
-          # NOTE: using the git CLI rather than `gh api git/{tags,refs}`.
-          # A git-push of a tag pointing at an existing commit produces a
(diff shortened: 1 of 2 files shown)


lczyk:
switching tag creation to `gh` seems to cause some breakages. investigating...

edit: so this is fun. rest apicreated tag refs fire the `Image.yaml` push event so they were re-firing the CI workflow :laughing: let me try `git`...

.github/workflows/Release.yaml
```diff
@@ -235,17 +235,60 @@ jobs:
...
+
+          # Tagger comes from git config at tag-creation time.
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+
+          # Passing TARGET_REF (itself an annotated tag) without peeling
+          # preserves the prior tag-of-tag chain back to the upload tag.
+          git tag -fa "${TAG}" -m "${MSG}" "${TARGET_REF}"
+
+          # `actions/checkout` installs a GITHUB_TOKEN-based extraheader
```
lczyk: this is hairy. i did not know about the `.extratoken` and i don't think i like it. i've written the full replace/restore config just for completeness here (e.g. restore is not actually needed at all in this one place, but then if someone adds a new step after, the auth leaks through `.extratoken`). i think in practice we should set `persist-credentials: false` on `actions/checkout` though (!!)

.github/workflows/Release.yaml
```diff
@@ -235,17 +235,60 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ secrets.ROCKSBOT_TOKEN }}
+          TAG: ${{ matrix.release-name }}
+        run: |
+          set -e
+          if [[ "$RUNNER_DEBUG" == "1" ]]; then
+            set -x
+          fi
+
+          # unmatched glob -> empty file list, release still created.
```
lczyk, line 281: this is here to match `softprops/action-gh-release` behaviour. do we want this to behave like this though..?

.github/workflows/Release.yaml
```diff
@@ -235,17 +238,51 @@ jobs:
...
+          # The tag's tagger field comes from git config below, so the tag
+          # object still records ROCKsBot regardless of the push token.
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+
+          # Passing TARGET_REF (itself an annotated tag) without peeling
+          # preserves the prior tag-of-tag chain back to the upload tag.
+          git tag -fa "${TAG}" -m "${MSG}" "${TARGET_REF}"
+
+          # NOTE: the tag push below relies on the GITHUB_TOKEN-based auth
```
lczyk, line 260: i've [tried](https://github.com/canonical/oci-factory/pull/1025#discussion_r3287990439) to make this work with ROCKSBOT_TOKEN but that's what caused the pathological recursive workflow blowup. ¡thank you! @zhijie-yang for fire fighting. anyway, looks like the only way we can push without re-triggering ourselves in turn. also see https://github.com/canonical/oci-factory/pull/1025/changes#diff-95fcfe303949acf55eb70ef4d76248b68599742ee910a3be1b985a21c73a3495R6-R12 -- defensive guard on Image.yaml. didn't help here because workflow picked up Image.yaml from main, not from PR.

.github/workflows/Release.yaml
lczyk: closing in favor of https://github.com/canonical/oci-factory/pull/1025#discussion_r3289506129

PR #1119, opened 2026-06-23
chore: update Valkey image

reneradoi:

---

## Description
Update Valkey image to newer version 7.2.12, and address security issues raised on the previous image.

### Related issues
Security issues raised by the ROCKsBot, e.g. https://github.com/canonical/charmed-valkey-rock/issues/41

### DIFF
--- a/oci/valkey/image.yaml
+++ b/oci/valkey/image.yaml
@@ -2,10 +2,10 @@ version: 1
 
 upload:
   - source: "canonical/charmed-valkey-rock"
-    commit: 916eff983d28129053093a560ff2e4a48e207ffc
+    commit: b347a0ecb3eb7e1c29d297e1cafcb6e46a66a0c9
     directory: .
     release:
-      7.2.11-24.04:
+      7.2.12-24.04:
         end-of-life: "2029-04-01T00:00:00Z"
         risks:
           - stable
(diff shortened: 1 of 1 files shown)


zhijie-yang (changes requested):
Please observe the CVE findings: https://github.com/canonical/oci-factory/actions/runs/28020636779/attempts/1#summary-82937249875

reneradoi:
Hi @zhijie-yang thank you for reviewing. I have updated the image to fix the remaining CVEs, could you please have another look at the PR? Thank you!

zhijie-yang:
After looking back at the build recipe for valkey again, I'd like to raise a question regarding the versioning of this rock: Since the `charmed-valkey` is pulled from `charmed-valkey/7/edge` as the [`stage-snap`](https://github.com/canonical/charmed-valkey-rock/blob/7-24.04/rockcraft.yaml#L37), the `version` specified in the `rockcraft.yaml` does not provide a source of truth as the channel moves. Shall we just remove the minor and patch versions from the `version` field in the `rockcraft.yaml` file, as well as the track name?

reneradoi:
Hi @zhijie-yang! We currently use the `version` field in `rockcraft.yaml` to gate releases, updating it with a PR to release a new rock when we have updated the snap. I agree, the solution is not ideal. As we are currently reworking our rock image repositories (we have multiple per product) to make more sense, we will take this into account. But for the current one, I'd like to not interfere with the other changes and keep it as is, if this ok for you?

zhijie-yang (approved):
noted. thanks for the explanation. let's deal with the version field in a future release.

PR #1138, opened 2026-07-01
chore: update ruby

alesancor1:
- [x] I have signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [x] I am following the [CONTRIBUTING](https://github.com/canonical/oci-factory/blob/main/CONTRIBUTING.md) guidelines

---

## Description
New arches for ruby. Now builds on s390x and ppc64

PR #1139, opened 2026-07-01
chore: update rabbitmq

alesancor1:

---

## Description
New arches for rabbitmq. Now builds on s390x and ppc64le

### DIFF
--- a/oci/rabbitmq/image.yaml
+++ b/oci/rabbitmq/image.yaml
@@ -2,7 +2,7 @@ version: 2
 upload:
   - source: canonical/rabbitmq-rock
     commit: f58aea30b3743816e75050a2b18b44f16175c0a7
-    directory: 4.0-26.04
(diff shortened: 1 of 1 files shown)


zhijie-yang (commented):
wow this is a very cautious move to kick everything back to the `edge` risk. Do we need this since this is just an addition of architectures? Since the build recipes (except for the "architecture" field) didn't change, why can't we put them directly to candidate, and have a bump to stable later?

alesancor1:
Just abiding by best practices here... A new architecture is essentially a new image so it should go to `edge` first and follow the usual lifecycle of promotions

zhijie-yang (approved):
Trigger looks good to me. Since there is not an automation to promote the risks to stable, we should a least have an issue/Jira ticket to track this. Same comment applies to #1138 .

PR #1140, opened 2026-07-01
feat(ci): use swift container from PS7

zhijie-yang:

---

## Description

OCI Factory now uses the OpenStack Swift container instance in PS7.

### Related issues

---

*Picture of a cool rock:*


src/uploads/requirements.sh
```diff
@@ -3,4 +3,4 @@
 sudo apt update
 
 # Get ubuntu-distro-info
-sudo apt install -y distro-info jq
\ No newline at end of file
+sudo apt install -y distro-info jq zip
```
alesancor1, line 6: qq: what do we need zip for?

src/image/requirements.sh
```diff
@@ -15,3 +15,8 @@ popd
 
 sudo apt update
 sudo apt install -y distro-info
+
+# Install aws cli if not already installed
+if ! command -v aws &> /dev/null; then
+    sudo snap install aws-cli --classic
+fi
```
alesancor1, line 22: qq: what do we need it for?

.github/workflows/_Auto-updates.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
 
   check-auto-updates:
     name: Check for auto-updates based on ${{ matrix.base }}
-    runs-on: ubuntu-22.04
+    runs-on: self-hosted-linux-amd64-noble-private-endpoint-small
```
alesancor1, line 37: Note OCI-Factory is a public repo, can we still use this?

alesancor1 (commented):
Changes LGTM, just couple of questions.

.github/workflows/_Auto-updates.yaml
zhijie-yang: Yes, and this is a must --- Otherwise we cannot access the Swift container in PS7.

src/image/requirements.sh
zhijie-yang: When publishing, the aws cli is called to generate the API key that can be consumed by skopeo. See `publish_to_aws_ecr_public()` in `src/image/tag_and_publish.sh`.

zhijie-yang: And the `private-endpoint` runners don't have it installed by default.

src/uploads/requirements.sh
zhijie-yang: To pack the sboms. See step `generate-sboms` in job `upload`. The GH-hosted and non-private-endpoint runners have it installed by default.

alesancor1 (approved):

src/uploads/oci_registry_upload.py
```diff
@@ -30,7 +30,14 @@ def parse_args():
 
 def main():
     args = parse_args()
-    base_cmd = ["skopeo", "copy", "--preserve-digests", args.source_uri]
+    # --insecure-policy is required on private-endpoint runners.
```
cjdcordeiro, line 33: 🤔

cjdcordeiro (commented):
it's in draft, but so far lgtm

src/uploads/oci_registry_upload.py
zhijie-yang: The private runner does not provide a default `policy.json`, which is provided by `golang-github-containers-common`. However, even in that package, the provided `/etc/containers/policy.json` is just

```
{
    "default": [
        {
            "type": "insecureAcceptAnything"
        }
    ]
}
```

It should be acceptable that we explicitly allow insecure accept anything, instead of using such a package and implicitly allowing insecure accept anything.

Happy to change to the alternative using `sudo apt install golang-github-containers-common` in the `requirements.sh`. WDYT?

zhijie-yang: I'll merge the PR. Feel free to put up an issue if you find this needs improvements.

PR #1153, opened 2026-07-17
feat: pro rock in ACR

zhijie-yang:
Jobs to be done: Rebuild, Continuous scanning

---

## Description

Add support for building and releasing Ubuntu Pro Rocks to a private Azure Container Registry.

### Changes

- Extend the image trigger schema with `upload[*].pro.services` (user-facing) and `pro-release` (no user-facing).
- Validate supported, unique Ubuntu Pro services.
- Build Pro images using centrally managed Pro credentials.
- Encrypt Pro OCI archives with GPG while stored as workflow artifacts.
- Publish Pro images exclusively to ACR and track them in `_pro_releases.json`. (no uploads to GHCR)
- Keep public GHCR, Docker Hub, ECR, GitHub Release, and documentation flows unchanged.
- Add Pro-aware rebuild, vulnerability-scan, release, and continuous-testing workflows.
- Cache encrypted archives and decrypt them only in ephemeral test jobs.
- Add unit, integration, workflow-contract, and publishing tests.
- Document Pro image configuration and current limitations.

### Limitations

- Pro support is currently a preview.
- Existing Pro revisions cannot be promoted using `pro-release`; releases must be done via a rebuild.
- Pro releases do not update registry documentation, as ACR does not have a WebUI for the docs.

### Testing

All PR checks pass, including `pytest`, `bats`, workflow contract tests, and mock-rock build, test, and release workflows.

### Related issues

---

*Picture of a cool rock:*


.github/workflows/Image.yaml
```diff
@@ -208,9 +212,11 @@ jobs:
     uses: ./.github/workflows/Test-Rock.yaml
     with:
       oci-archive-name: "${{ matrix.name }}_${{ matrix.commit }}_${{ matrix.dir_identifier }}"
+      encrypted-artifact: ${{ matrix.pro-services != '' }}
```
zhijie-yang: Check https://github.com/canonical/rocks-template/blob/main/.github/workflows/image.yaml#L53 and the test workflow. This should be handled without using a new input.

zhijie-yang: Done in [ee53cfd](https://github.com/canonical/oci-factory/pull/1153/commits/ee53cfd99b01e019422ad98e54d6b91e360558e0)

.github/workflows/Image.yaml
```diff
@@ -487,16 +499,26 @@ jobs:
...
 
+      - name: Encrypt pro artifact
+        if: ${{ matrix.pro-services != '' }}
+        uses: ./.github/actions/crypt-artifact
+        with:
+          mode: encrypt
+          input-path: ${{ steps.rename-oci-archive.outputs.name }}
+          passphrase: ${{ secrets.ROCKS_PRO_ARTIFACT_PASSPHRASE }}
+          output-path: ${{ steps.rename-oci-archive.outputs.name }}.gpg
+
```
zhijie-yang, line 516: See if this can be optimized, we can just rename the downloaded artifact, and choose to keep the encrypted artifact during the decryption.

.github/workflows/Release.yaml
```diff
@@ -83,8 +83,14 @@ jobs:
...
     needs: [validate-push-release-request]
+    concurrency:
+      group: release-${{ inputs.oci-image-name }}
+      cancel-in-progress: false
     outputs:
-      gh-releases-matrix: ${{ steps.release-image.outputs.gh-releases-matrix }}
+      gh-releases-matrix: ${{ steps.release-image.outputs.gh-releases-matrix || '{"include":[]}' }}
+      has-github-releases: ${{ steps.release-image.outputs.has-github-releases }}
+      is-public-release: ${{ steps.release-type.outputs.is-public-release }}
+      is-pro-release: ${{ steps.release-type.outputs.is-pro-release }}
```
zhijie-yang, line 93: Consider refactoring these two boolean names to `contains-x-release` to reduce the ambiguity.

.github/workflows/Test-Rock.yaml
zhijie-yang, line 1: Revisit this to see if we can avoid any changes to this established workflow.

.github/workflows/Image.yaml
```diff
@@ -197,7 +197,11 @@ jobs:
       rock-repo-commit: ${{ matrix.commit }}
       rockfile-directory: ${{ matrix.directory }}
       lpci-fallback: true
-    secrets: inherit
+      pro-services: ${{ matrix.pro-services }}
+    secrets:
+      source-github-token: ${{ secrets.ROCKSBOT_TOKEN }}
+      pro-token: ${{ secrets.ROCKS_PRO_TOKEN }}
+      pro-artifact-passphrase: ${{ secrets.ROCKS_PRO_ARTIFACT_PASSPHRASE }}
```
lczyk:
```suggestion
      pro-artifact-passphrase: ${{ matrix.pro-services != '' && secrets.ROCKS_PRO_ARTIFACT_PASSPHRASE || '' }}
```

README.md
```diff
@@ -513,6 +552,7 @@ needed.
 | Property | Required | Type | Description |
 |---|---|---|---|
 |`oci-archive-name`| True | str | Artifact name to download for testing. |
+|`encrypted-artifact`| False | bool | Whether the OCI archive artifact is encrypted. Defaults to `false`. Supplying `pro-artifact-passphrase` also selects encrypted-artifact handling. |
```
lczyk: that input does not exist...

README.md
```diff
@@ -230,14 +236,44 @@ upload:
...
+Ubuntu Pro image credentials are centrally managed by OCI Factory. Maintainers
+only select the services to enable through `upload[*].pro.services`.
+
+Pro image archives are encrypted while being exchanged and retained as GitHub
+workflow artifacts. SBOMs, vulnerability reports, and build metadata are not
+encrypted.
+
+Pro support is currently a preview with these limitations:
+
+- continuous post-release vulnerability scanning does not support Pro images,
```
lczyk:
```suggestion
```

surely it does now?

src/image/merge_release_info.py
```diff
@@ -44,6 +48,20 @@ def backfill_higher_risks(channels: dict) -> None:
...
+def _assemble_pro_track(_track: str, _pro_services: list[str]) -> str:
+    """Assembles the track name with the pro services."""
+    # Remove any esm- services from the pro services to form the tag,
+    # as they are excluded from the pro track names.
+    _pro_services = [ps for ps in _pro_services if not ps.startswith("esm-")]
+
+    if not _pro_services:
+        return _track
+
+    version, base = _track.rsplit("-", 1)
```
lczyk, line 60: this throws if _track does not have `-` in it

tools/workflow-engine/charms/temporal-worker/oci_factory/activities/find_images_to_update.py
```diff
@@ -135,62 +196,90 @@ def trigger_image_rebuild():
...
-                    timezone.utc
-                ).strftime("%Y-%m-%dT%H:%M:%SZ"):
+        if pro:
+            original_track = build_metadata["track"]
+            visible_services = [
+                service for service in services if not service.startswith("esm-")
+            ]
+            expected_track = original_track
+            if visible_services:
+                version, base_release = original_track.rsplit("-", 1)
```
lczyk, line 206: same comment as https://github.com/canonical/oci-factory/pull/1153/changes#r3648767498

.github/workflows/Build-Rock.yaml
lczyk: i think this might not strip path quite correctly.  `%%` strips from the **first** dot, not last, so it can catch a dot in the middle of the version number

lczyk, line 272: this workflow reads secrets which are not explicitly declared in `on.workflow_call.secrets`. this used to work because of `secrets: inherit` but now we need a bit more plumbing

lczyk (changes requested):
as discussed on 1-1, overall approach looks good. a couple of drive-bys but i personally don't mind. i found a small pile of things to look at though:

- worth failing fast in the matrix on an `upload[*]` entry with `pro:` but no `release:`
- pro build metadata in Swift records `digest: ""` (the GHCR push is skipped), so a Pro revision has no content identity anywhere once the artifact expires.   
- pro archives are uploaded with default retention (90 day). `retention-days: 1` ??
- `Create Git tag` runs for Pro too. just flagging.
- `upload-rock/action.yml` still decrypts with a `<name>*` glob while everything else moved to `<name>.gpg`.
- `docker` is unpinned in `tests/etc/requirements.txt` while everything else there is pinned. 

and inline comments ofc

.github/workflows/Build-Rock.yaml
zhijie-yang: We already ditched LP builds. They code won't be called and is expected to be removed soon.

src/image/merge_release_info.py
zhijie-yang: It's expected. The track would be considered as ill-formed otherwise. This is also an established pattern in OCI-Factory, dating back to more than 3 years ago:

https://github.com/canonical/oci-factory/blob/9d4267cd6edba309b7cf80366e2fadf7f5eec56b/src/docs/generate_oci_doc_yaml.py#L318

.github/workflows/Build-Rock.yaml
zhijie-yang: Interesting. This seems worthy of a fix @alesancor1

src/image/merge_release_info.py
lczyk: sure. just pointing it out. it fails loudly anyway so much less of an issue that it would be otherwise

lczyk (approved):
https://github.com/canonical/oci-factory/pull/1153#discussion_r3648869736 was really my only blocker. everything else is flags / nice-to-have's so +1 from me

lczyk (changes requested):
```diff
diff --git i/src/uploads/oci_registry_upload.py w/src/uploads/oci_registry_upload.py
index 04a7d944..a2611dda 100755
--- i/src/uploads/oci_registry_upload.py
+++ w/src/uploads/oci_registry_upload.py
@@ -40,6 +40,10 @@ def main():
             "--rm",
             "-v",
             f"{tmp_dir}:/skopeo-auth:ro",
+            "-v",
+            f"{os.getcwd()}:/workdir:ro",
+            "-w",
+            "/workdir",
             os.environ.get("SKOPEO_IMAGE", DEFAULT_SKOPEO_IMAGE),
             "--insecure-policy",
             "copy",
```

damn. missed it. smth like this ^ to fix the upload i think

lczyk (approved):
yep. https://github.com/canonical/oci-factory/pull/1153/commits/f2a3f98b94e1e194b13eee9fe12b70693bcfa505 is even better. *now* approve ;)

.github/workflows/Build-Rock.yaml
zhijie-yang: Fixed in 54f62604

.github/actions/commit-releases-json/action.yaml
```diff
@@ -46,7 +46,20 @@ runs:
...
+            release_files+=("$path")
+          fi
+        done
+
+        if [[ ${#release_files[@]} -eq 0 ]]; then
+          echo "No release files found for ${{ inputs.image-name }}" >&2
+          exit 1
+        fi
+
+        git add -- "${release_files[@]}"
```
alesancor1, line 62: This is a public repository, so please note that committing the `_pro_releases.json` file would at least expose a list of the images we have in ACR requiring a `pro` subscription.

While it's just listing the images, this would be allowing an operation that is usually protected behind authentication & authorization. We need to ensure we want to allow this.

---

Potential workaround: bring and commit the pro releases files to a private repository, which you can access with a token.

.github/workflows/_Test-OCI-Factory.yaml
alesancor1: If you remove `continue-on-error` then you have to make this run regardless:

```suggestion
        if: always()
```

For reference, this is what we do in the rocks-template-actions repository. See [here](https://github.com/canonical/rocks-template-actions/blob/main/.github/workflows/_Test-Actions.yaml)

alesancor1: Same as above

```suggestion
        if: always()
```

.github/workflows/Build-Rock.yaml
```diff
@@ -328,7 +333,7 @@ jobs:
             ./.github/actions/crypt-artifact/crypt-artifact.sh decrypt \
               -i "$rock" \
               -p "$ARTIFACT_PASSPHRASE" \
-              -o "${rock%%.*}.rock"
+              -o "${rock%.gpg}"
```
alesancor1: Have you tested it? Note that by default if you don't provide `-o` the `crypt-artifact.sh` script will set the output to `$rock%.gpg` so if that works the same then you may want to remove `-o` completely.

.github/workflows/Test-Rock.yaml
alesancor1: These changes seem more like a drive-by to me. Afaict they are not really required for any of the workflows to run correctly right? Since the inputs and outputs remain the same.

If so, I'd move this to a separate refactor PR.

alesancor1 (changes requested):
First pass

Overall the code looks good, I can see it's quite polished already after a couple iterations with @lczyk (thanks!).

For now, just a couple of nits and a bigger concern, check below:

.github/actions/commit-releases-json/action.yaml
zhijie-yang: There is no point moving this out. OCI-Factory is a public repository, and thus even though these info are committed to a private repo, the name of the image will still be there. Plus, that these info are also available in the image trigger and workflow logs, there is no need to make such a protection. Any illegal accesses to ACR should be blocked by ACR's authentication.

.github/workflows/Test-Rock.yaml
zhijie-yang: These changes aren't a drive-by refactor — they're required for Pro support, which is why the inputs/outputs are unchanged (the behavior change is internal to how Pro artifacts are handled).

Previously `configure-tests` unpacked the rock once and cached the **unpacked OCI layout**, which every downstream test job restored. For Pro images that would mean the **decrypted** image content sits in the shared, run-scoped cache — defeating the reason we GPG-encrypt the artifact in the first place.

To keep the property "cache encrypted archives, decrypt only in ephemeral test jobs", the restructuring:

- caches the raw archive (`<name>` / `<name>.gpg`) instead of the unpacked layout, and
- moves the `Decrypt Rock` + `Unpack Rock` steps out of the shared job into each individual test job, so decryption happens only inside the short-lived job that needs it and the plaintext is never persisted to the cache.

Public (non-Pro) images are unaffected: the `Decrypt Rock` step is gated on `ARTIFACT_PASSPHRASE != ''`, so for them it's a no-op and the unpack behaves as before.

Given that, I'd prefer to keep it in this PR rather than split it, since the Pro test flow doesn't work correctly without it. Happy to revisit if you still see a piece here that's genuinely independent of Pro.

alesancor1: Yes, this is essentially what I flagged back in: https://github.com/canonical/oci-factory/pull/706/changes#r2623388879 but wasn't ever addressed. Happy to keep it in this PR 👍

alesancor1 (approved):
Happy with the current state of this, thanks!

Is there any chance you could provide a test run before merging?

zhijie-yang:
It worked happily after the merge of main. See workflow run: https://github.com/canonical/oci-factory/actions/runs/33501795526/job/99895265422#step:6:5. Merging this.

